PR #3, opened 2026-01-27
fix(plugins): resolve plugin installation and restore pipeline issues

nghorbani:


examples/adaclip/concrete_adaclip_gradient_training.py
```diff
@@ -56,18 +55,18 @@ def main(cfg: DictConfig) -> None:
...
     logger.info("Loading AdaCLIP plugin from GitLab repository...")
-    
+
     # Create a NodeRegistry instance for plugin loading
     registry = NodeRegistry()
     registry.load_plugin(
         name="adaclip",
         config={
-            "repo": "git@gitlab.cubert.local:cubert/cuvis-ai-adaclip.git",
+            "repo": "git@github.com:cubert-hyperspectral/cuvis-ai-adaclip.git",
```
birkholz-cubert: I would suggest that we should by default pull via https due to plugins probably being public repos anyways.

i could not get the install by git to work, the cloned repo seems to not be a git repository for some reason. Local path register works fine.

nghorbani: could you try: https://github.com/cubert-hyperspectral/cuvis-ai-adaclip.git

nghorbani: i changed the imports from ssh to https. dropped hash and branch ref. renamed ref to tag. only supporting tag or local custom nodes

birkholz-cubert (approved):

PR #20, opened 2026-04-02
Fix AP metric state leak and add AdaCLIP finetuning workflow (RGB/CIR)

rajanish4:
## Summary
- Fixes `AnomalyDetectionMetrics` AP state accumulation to prevent per-batch growth/memory pressure during validation.
- Adds `AdaCLIPFocalDiceLoss` and focused unit tests covering per-layer path, fallback path, finite scalar output, gradient flow, and loss weighting.
- Adds AdaCLIP finetuning workflow wiring (RGB/CIR scripts + trainrun/plugin config updates).
## Test plan
- [x] `uv run pytest`
  - Result: 836 passed, 16 skipped
- [x] `uv run pytest tests/training/test_losses.py tests/training/test_metrics.py -q`
  - Result: all passed
## Notes
- PR base is set to `feature/anish/all-5489-support-multi-file-lentils-datasets-via-splits-csv-and-refactor-data-loading` so only the intended 11-file delta is included.


nghorbani:
## Review Comments

### 1. `json_reader.py` / `json_writer.py` — should be removed
These were already combined into `json_file.py` on `main`. This branch re-splits them into separate files, which conflicts with the consolidation. These should be removed and the existing `json_file.py` from main should be used instead.

### 2. SAM3 changes — should be reverted to main
The branch adds 3 new SAM3 pipeline configs (`sam3_learned_projection.yaml`, `sam3_naive_false_rgb.yaml`, `sam3_with_spectral_signatures.yaml`) and 2 new example scripts (`sam3_hsi_tracker.py`, `sam3_video_tracker.py`) that don't belong in this PR. The SAM3 configs and examples from main should remain as-is.

### 3. `examples/object_tracking/README.md` — should be reverted to main
The tracking README was heavily rewritten (simplified prompts, removed scheduled features, new detection spec). These changes are out of scope for this PR and should be reverted to main's version.

### 4. Video & Visualization changes — should be reverted
Changes to `cuvis_ai/node/video.py` (removed `overlay_title`, `_darken_region`, `_draw_title_overlay`) and `cuvis_ai/utils/vis_helpers.py` (removed `object_color`, `render_multi_object_overlay`) are tracking infrastructure changes that shouldn't be in this AdaCLIP finetuning PR.

### 5. `cuvis_ai/training/schemas.py` — should live in `cuvis-ai-schemas`
`MultiFileDataConfig` and `MultiFileTrainRunConfig` belong in the `cuvis-ai-schemas` package, not in `cuvis-ai` itself. The file header even says "Temporary workaround until cuvis-ai-schemas DataConfig is extended upstream." This should be moved upstream rather than shipped as a local override.

### 6. `cuvis_ai/node/normalization.py` — `max_initialization_frames` is NOT a duplication
Verified that main does **not** have this parameter. The change adds a genuine new feature (frame-limited init) plus a refactor from list accumulation to running min/max. This is fine to keep, but worth confirming it doesn't overlap with any planned fixed-step normalization in the upstream schemas.

cuvis_ai/node/losses.py
```diff
@@ -840,6 +840,215 @@ def forward(self, rgb: Tensor, mask: Tensor, **_: Any) -> dict[str, Tensor]:
         return {"loss": loss}
 
 
+class AdaCLIPFocalDiceLoss(LossNode):
```
nghorbani: **Required before merge:** please move this class (and the two finetune examples in `examples/adaclip/`) into the cubert-hyperspectral/cuvis-ai-adaclip repo. It's tightly coupled to `AdaCLIPDetector`'s `per_layer_scores` / `image_score_2ch` outputs, has no other consumers, and is paper-faithful to AdaCLIP's `method/trainer.py` — it belongs with the plugin. We don't ship Anomalib/Dinomaly/RT-DETR losses in cuvis-ai either; same pattern applies. The class only imports from cuvis-ai-core (`LossNode`, `PortSpec`), not from cuvis-ai itself, so the port is mechanical. Examples can depend on cuvis-ai as an example-only requirement. Add corresponding tests on the plugin side. What stays in cuvis-ai: the plugin manifest entry and (if useful) the trainrun YAML.

configs/plugins/adaclip.yaml
```diff
@@ -8,8 +8,8 @@
 
 plugins:
   adaclip:
-    # AdaCLIP-based anomaly detection with advanced band selection
-    repo: "https://github.com/cubert-hyperspectral/cuvis-ai-adaclip.git"
-    tag: "v0.1.2"
+    # Local development path (feature/anish/adaclip-finetuning branch)
+    # For production, use: repo: "https://github.com/cubert-hyperspectral/cuvis-ai-adaclip.git"
+    path: "/home/dev/anish/cuvis-ai-adaclip"
```
nghorbani: **Required before merge:** revert this. The hardcoded `/home/dev/anish/cuvis-ai-adaclip` breaks CI and any other developer's environment. Once cubert-hyperspectral/cuvis-ai-adaclip#9 is tagged `v0.2.0`, restore `repo:` + `tag: "v0.2.0"` here. Same with the `cd /home/dev/anish/cuvis-ai` lines in the example docstrings — generic placeholders please.

cuvis_ai/node/metrics.py
```diff
@@ -223,14 +227,19 @@ def forward(
         ]
 
         if logits is not None:
+            # Prevent BinaryAveragePrecision from accumulating every pixel of every
+            # validation batch across all epochs (torchmetrics forward() updates state).
+            self.average_precision_metric.reset()
```
nghorbani: The reset-before/reset-after hygiene is correct, and `test_average_precision_state_does_not_accumulate_batches` is a solid regression test.

One note for the CHANGELOG: previously AP was effectively *cumulative* across all val batches because state was never reset. After this fix it's strictly per-batch — matches the other metrics in this node, but anyone comparing AP curves across runs will see the values shift. Worth flagging as a behavioral note, not just a memory fix. The current diff's CHANGELOG entries are SAM3/ByteTrack items from the base branch merge; this PR's actual changes (AP fix, AdaCLIP loss) aren't represented yet.

cuvis_ai/node/losses.py
```diff
@@ -840,6 +840,215 @@ def forward(self, rgb: Tensor, mask: Tensor, **_: Any) -> dict[str, Tensor]:
...
+                is_anomaly = (gt_binary.flatten(1).sum(dim=1) > 0).long()  # [B]
+                cls_loss = self._mc_focal(
+                    image_score_2ch.unsqueeze(-1),  # [B, 2, 1]
+                    is_anomaly.unsqueeze(-1),  # [B, 1]
+                )
+                loss = loss + self.image_loss_weight * cls_loss
+        else:
+            # ── Fallback: aggregated binary loss (val/test) ───────────
+            pred = predictions.squeeze(-1)  # [B, H, W]
+            p_min, p_max = pred.min(), pred.max()
```
nghorbani: Min-max rescaling inside the loss makes gradients non-stationary because the scale depends on each batch's min/max. If `predictions` is supposed to be in [0,1], drop the rescale and document the contract on the port spec; if it might be logits, replace with `pred.sigmoid()`. Worth resolving when this code moves to the plugin repo.

cuvis_ai/node/losses.py
```diff
@@ -840,6 +840,215 @@ def forward(self, rgb: Tensor, mask: Tensor, **_: Any) -> dict[str, Tensor]:
...
+        predictions: Tensor,
+        targets: Tensor,
+        per_layer_scores: Tensor | None = None,
+        image_score_2ch: Tensor | None = None,
+        **_: Any,
+    ) -> dict[str, Tensor]:
+        gt = targets.float().squeeze(-1)  # [B, H, W]
+        gt_binary = gt.clone()
+        gt_binary[gt_binary > 0.5] = 1
+        gt_binary[gt_binary <= 0.5] = 0
```
nghorbani: These two lines simplify to `gt_binary = (gt > 0.5).float()`.

cuvis_ai/node/losses.py
```diff
@@ -840,6 +840,215 @@ def forward(self, rgb: Tensor, mask: Tensor, **_: Any) -> dict[str, Tensor]:
...
+        gamma = self.focal_gamma
+        smooth = self.focal_smooth
+
+        if logit.dim() > 2:
+            logit = logit.view(logit.size(0), logit.size(1), -1)  # [B, C, HW]
+            logit = logit.permute(0, 2, 1).contiguous()  # [B, HW, C]
+            logit = logit.view(-1, logit.size(-1))  # [N, C]
+        target = target.reshape(-1, 1).long()  # [N, 1]
+
+        alpha = torch.ones(num_class, 1, device=logit.device, dtype=logit.dtype)
```
nghorbani: `alpha = torch.ones(num_class, 1)` makes alpha uniform, which means the focal weight is the same for both classes — i.e. the alpha indexing a few lines below is effectively a no-op. Either drop the alpha math or expose alpha as a constructor parameter for class imbalance.

examples/adaclip/finetune_adaclip_cir.py
```diff
@@ -0,0 +1,243 @@
+"""Fine-tune AdaCLIP adapter layers with CIR (Color Infrared) band selection.
```
nghorbani: These two finetune scripts (`finetune_adaclip_rgb.py` and `finetune_adaclip_cir.py`) are about 95% duplicated — only the selector differs. Either extract a `build_finetune_pipeline(selector, ...)` helper or use Hydra to switch the selector. They'll rot the moment one side is updated and the other isn't. Worth resolving when this code moves to the plugin repo.

configs/pipeline/anomaly/adaclip/adaclip_cir_false_color.yaml
```diff
@@ -16,7 +16,7 @@ nodes:
   hparams:
     weight_name: pretrained_all
     backbone: ViT-L-14-336
-    prompt_text: 'normal: lentils, anomaly: stones'
+    prompt_text: '' # 'normal: lentils, anomaly: stones'
```
nghorbani: `prompt_text: ''` with the original commented out — looks WIP. Confirm this is intentional? AdaCLIP's text branch is meaningful for the contrastive signal; an empty prompt may not be what you want at inference.

docs/user-guide/installation.md
```diff
@@ -47,10 +47,8 @@ conda install -c conda-forge ffmpeg
 # macOS
 brew install ffmpeg
 
-# Windows (scoop) — use the *shared* build so torchcodec can find the DLLs
-scoop install ffmpeg-shared
-# Then add to PATH (Git Bash / MSYS2):
-export PATH="/c/Users/$USER/scoop/apps/ffmpeg-shared/current/bin:$PATH"
+# Windows — download a release build and add the bin/ folder to PATH
```
nghorbani: The previous block had concrete scoop instructions including the shared-DLL caveat (`ffmpeg-shared`) that torchcodec needs to find the DLLs. Replacing it with a generic ffmpeg.org link will leave Windows users stuck on the very issue the original block was solving. Please restore the platform-specific block.

nghorbani (commented):
This bundles two unrelated changes — the AP metric fix is shippable today; the AdaCLIP finetuning workflow needs rework before it can land here. Please consider splitting them.

I've left line-anchored comments inline. Two of them are required before merge (move the plugin-specific code into cuvis-ai-adaclip, revert the hardcoded path); the rest are smaller observations.

Suggested order:

1. Land cubert-hyperspectral/cuvis-ai-adaclip#9 + tag `v0.2.0` (review on that PR is separate).
2. Move the AdaCLIP-specific code (loss + finetune examples) out of this PR into the plugin repo — see comment on `cuvis_ai/node/losses.py`.
3. Restore the plugin manifest to `repo:` + `tag: v0.2.0`.
4. Either split out the AP fix as its own PR (clean cherry-pick) or keep bundled but make the CHANGELOG reflect both changes.

rajanish4:
Thanks for the detailed review. I applied a split strategy and reduced this PR to AP-fix scope only.

Current PR scope now contains only:
- cuvis_ai/node/metrics.py AP state reset fix
- tests/training/test_metrics.py regression coverage for non-accumulating AP state
- CHANGELOG.md behavioral note (AP now per-batch instead of cumulative across validation batches)

Required comments:
1) Move AdaCLIP loss/examples out of cuvis-ai: addressed by removing those additions from this PR and tracking in plugin follow-up.
2) Revert hardcoded adaclip local path: addressed by removing this change from PR scope.

Advisory comments handled in this PR:
- Added changelog note for AP behavior shift.
- Removed docs/config changes that were out-of-scope for this PR.

Advisory comments deferred to plugin follow-up (where code now belongs):
- loss fallback min-max gradient concern
- gt_binary simplification
- focal alpha no-op observation
- RGB/CIR finetune duplication

Validation run:
- uv run pytest tests/training/test_metrics.py -k average_precision_state_does_not_accumulate_batches -q
- uv run pytest tests/training/test_metrics.py -q
- uv run ruff format --check cuvis_ai/node/metrics.py tests/training/test_metrics.py

nghorbani (approved):

PR #37, opened 2026-05-29
node library: plugins block, inline node catalogs, dep-compat CI

nghorbani:
This PR lands the node-library + configs slice of the plugin-loader refactor: pipelines declare the plugin set they depend on, each plugin manifest carries an inline node catalog, and the dependency-compatibility CI lands. It backfills a bare-name `plugins:` block into every pipeline YAML, regenerates the manifests with inline `CatalogNodeEntry` catalogs in the single-`PortSpec`-per-port shape, adds the `dep_compat` and `registry_compat` workflows, and ships pipeline render + smoke tooling. The goal is for pipelines to name exactly what they depend on so the new core resolver/orchestrator can compose the right environment, with CI keeping the plugin and core dependency floors compatible.

### Pipelines
- A `plugins:` block (bare-name list) backfilled into every pipeline YAML.
- `scripts/backfill_pipeline_plugins.py` seeds the field, delegating to core's
  `reorder_pipeline_with_plugins` helper.

### Plugin manifests (`configs/plugins/*.yaml`)
- Each manifest's `provides:` carries the inline node catalog (`CatalogNodeEntry`).
- `turbovec` manifest added; `package_name` overrides declared; `PluginManifest`
  imported from `cuvis_ai_schemas.plugin`.

### CI
- `.github/workflows/dep_compat.yml` (host-floor check) and `registry_compat.yml`
  (plugin-vs-core check) added; dependency floors pinned.

### Nodes / pipelines
- SoftChannelSelector device-safe statistical init + gRPC train-flow fix.
- Monitor fan-in inputs marked variadic; manifests regenerated.

### Tooling / docs
- Transparent PNG renders of every pipeline yaml + smoke / render reports.
- `scripts/` made a PEP 420 namespace package; Windows orchestrator smoke unblocked.
- Plugin loading documented via `--plugins-dir`. CLAUDE.md added.

### Landing (release order)
Depends on the new cuvis-ai-core and cuvis-ai-schemas contracts. Lands last in
the chain, after cuvis-ai-core **0.7.0** and cuvis-ai-schemas **0.5.0** publish:
the `cuvis-ai-core` (`>=0.7.0`) and `cuvis-ai-schemas[full]` (`>=0.5.0`) floors
are bumped and the local editable `[tool.uv.sources]` overrides dropped so CI
resolves both from PyPI.


scripts/smoke_22_yamls.py
```diff
@@ -0,0 +1,232 @@
+"""One-shot smoke for the orchestrator path against every pipeline YAML.
```
birkholz-cubert: If you want this script to be public, maybe give it a more telling name

birkholz-cubert (approved):

PR #39, opened 2026-06-08
feat(node): generalize FixedWavelengthSelector to n-channel output

rajanish4:
## Summary

- Relaxes `ChannelSelectorBase.OUTPUT_SPECS["rgb_image"].shape` from `(-1,-1,-1,3)` to `(-1,-1,-1,-1)` so the port-spec validator accepts any channel count
- Generalizes `FixedWavelengthSelector.target_wavelengths` type annotation from `tuple[float, float, float]` to `tuple[float, ...]` with `len >= 1` validation
- Gates `_compose_rgb` (running bounds + gamma normalization) behind `normalize_output=True AND n==3`; for `n != 3` bands are stacked raw with a warning if `normalize_output=True`

**Fully backward compatible** — existing 3-channel callers are unaffected (verified by tests).

## Motivation

The `cuvis-ai-dinomaly` plugin needed a 6-channel band selector (VIS 625/550/450 nm + SWIR 1450/1200/1050 nm) for the bedding anomaly detection pilot. The upstream `FixedWavelengthSelector` rejected n≠3 output at the port-spec layer (`shape=(-1,-1,-1,3)`), forcing a plugin-local workaround (`FixedHyperspectralSelector`). This change removes the need for that workaround.

Closes: ALL-5779

## Test plan

- `tests/node/test_fixed_wavelength_selector.py` (16 new tests):
  - [ ] OUTPUT_SPECS shape is `(-1,-1,-1,-1)` on base class and selector
  - [ ] 3-channel default: shape, normalize path, band_info fields, nearest-band pick
  - [ ] 6-channel: shape `[B,H,W,6]`, dtype, band_info length, exact band values
  - [ ] Single-channel: shape `[B,H,W,1]`
  - [ ] `normalize_output=True` warning + raw fallback for n≠3
  - [ ] Constructor: empty tuple raises, bad type raises, int→float coercion
- 788 existing tests pass (`uv run pytest tests/ -q --ignore=tests/training --ignore=tests/anomaly`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,12 +117,7 @@ jobs:
 
       - name: Run pip-audit for vulnerabilities
         run: |
-          uv run pip-audit --desc on --skip-editable \
-            --ignore-vuln PYSEC-2022-42969 \
-            --ignore-vuln CVE-2025-71176 \
-            --ignore-vuln CVE-2026-34993 \
-            --ignore-vuln CVE-2026-47265 \
-            --ignore-vuln PYSEC-2026-196
+          uv run pip-audit --desc on --skip-editable --ignore-vuln PYSEC-2022-42969 --ignore-vuln CVE-2025-71176
 
       - name: Run detect-secrets
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,7 +117,12 @@ jobs:
 
       - name: Run pip-audit for vulnerabilities
         run: |
-          uv run pip-audit --desc on --skip-editable --ignore-vuln PYSEC-2022-42969 --ignore-vuln CVE-2025-71176
+          uv run pip-audit --desc on --skip-editable \
+            --ignore-vuln PYSEC-2022-42969 \
+            --ignore-vuln CVE-2025-71176 \
+            --ignore-vuln CVE-2026-34993 \
+            --ignore-vuln CVE-2026-47265 \
+            --ignore-vuln PYSEC-2026-196
 
       - name: Run detect-secrets
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
         with:
           fetch-depth: 0  # Full history for setuptools-scm
 
@@ -72,7 +72,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
 
       - name: Set up Python
         uses: actions/setup-python@v6
@@ -102,7 +102,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
 
       - name: Set up Python
         uses: actions/setup-python@v6
@@ -135,7 +135,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
 
       - name: Set up Python
         uses: actions/setup-python@v6
@@ -160,7 +160,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
         with:
           fetch-depth: 0
 
@@ -202,7 +202,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
         with:
           fetch-depth: 0  # Full history for setuptools-scm
 
--- a/cuvis_ai/node/channel_selector.py
+++ b/cuvis_ai/node/channel_selector.py
@@ -137,8 +137,10 @@ class ChannelSelectorBase(Node):
     OUTPUT_SPECS = {
         "rgb_image": PortSpec(
             dtype=torch.float32,
-            shape=(-1, -1, -1, 3),
-            description="Composed RGB image [B, H, W, 3] in 0-1 range",
+            shape=(-1, -1, -1, -1),
+            description="Composed image [B, H, W, C] in 0-1 range. "
+                        "C == 3 for standard RGB selectors; "
+                        "C == len(target_wavelengths) for FixedWavelengthSelector with n > 3.",
         ),
         "band_info": PortSpec(
             dtype=dict,
@@ -584,19 +586,25 @@ class NDVISelector(_NormalizedDifferenceIndexBase):
 
 
 class FixedWavelengthSelector(ChannelSelectorBase):
-    """Fixed wavelength band selection (e.g., 650, 550, 450 nm).
+    """Fixed wavelength band selection — picks the nearest band for each target wavelength.
 
-    Selects bands nearest to the specified target wavelengths for R, G, B channels.
-    This is the simplest band selection strategy that produces "true color-ish" images.
+    For the standard 3-channel case (default) this produces a "true color-ish" RGB image.
+    For ``n > 3`` target wavelengths the node stacks ``n`` bands into a
+    ``[B, H, W, n]`` output — useful for multi-channel hyperspectral models (e.g.
+    a 6-channel VIS+SWIR input to a Dinomaly detector).
 
     Parameters
     ----------
-    target_wavelengths : tuple[float, float, float]
-        Target wavelengths for R, G, B channels in nanometers.
-        Default: (650.0, 550.0, 450.0)
+    target_wavelengths : tuple[float, ...]
+        Target wavelengths in nanometers, in the order they should be stacked.
+        Must contain at least one wavelength.
+        Default: (650.0, 550.0, 450.0) — standard false-RGB.
     normalize_output : bool
-        If ``True`` (default), apply selector normalization to produce 0-1 RGB output.
-        If ``False``, return raw selected band values without normalization/gamma.
+        If ``True`` (default), apply selector normalization to produce a 0–1 output.
+        Normalization (running bounds + optional sRGB gamma) is only available for
+        the 3-channel case (``len(target_wavelengths) == 3``).  For ``n != 3`` with
+        ``normalize_output=True`` the bands are stacked without normalization and a
+        warning is emitted — pass ``normalize_output=False`` to suppress it.
     """
 
     _category = NodeCategory.TRANSFORM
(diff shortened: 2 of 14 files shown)


nghorbani (commented):
Reviewed the FixedWavelengthSelector generalization. The n-channel stacking on the plain `forward` path is correct and well covered by the new tests. The points below are in the normalization/statistical machinery the change does not generalize, plus the bundled pip-audit ignores. Inline notes sit on the specific lines; the full list is here for reference.

**Most important**

1. Statistical fit path is not guarded for `n != 3`. `ChannelSelectorBase.statistical_initialization` does `flat = raw_rgb.reshape(-1, 3)` and the `running_min`/`running_max` buffers are 3-element. The new gate only protects `forward`; the fit path is selected by `norm_mode == "statistical"` regardless of `normalize_output`. So `FixedWavelengthSelector(target_wavelengths=<6 wl>, norm_mode="statistical")` accumulates wrong stats (two channels mixed per row), and `n=4` raises `RuntimeError` during the fit phase. The same 3-element assumption breaks `running` mode if reached.
2. `band_info["normalized_output"]` reports `True` even when the bands were stacked raw (`n != 3` with `normalize_output=True`). A consumer that trusts this flag skips a normalization it still needs. Set it to `self.normalize_output and n == 3`.
3. The OUTPUT_SPECS shape was relaxed on the shared `ChannelSelectorBase`, which loosens the contract for every selector that inherits it (FastRGBSelector, RangeAverageFalseRGBSelector, the NDVI selectors), all of which always emit 3 channels. Consider overriding OUTPUT_SPECS on FixedWavelengthSelector only, leaving the base at `(-1, -1, -1, 3)` so the others keep the guard.
4. The `n != 3` warning fires on every `forward` call and imports loguru inside the function. In a video run that is one identical warning per frame. Both inputs are known at construction, so warn once in `__init__` and import loguru at module level (as the other node modules do).
5. The three new pip-audit ignores are unrelated to this change and are better handled in a deps PR. Two are aiohttp (CVE-2026-34993 deserialization, CVE-2026-47265 cookie leak), both fixed by `aiohttp >= 3.14.0`; the third is pip itself (PYSEC-2026-196 path traversal), fixed by `pip >= 26.1.2`. Prefer the upgrades over muting, and if anything must stay ignored, add a short comment per entry (package, reason, when to revisit).

**Minor**

6. The class docstring Ports section still says `shape (-1, -1, -1, 3)` and "0-1 range".
7. No test covers the `n != 3` statistical/running path, which is exactly the one that breaks (1).
8. `band_info["strategy"]` is hardcoded `"baseline_false_rgb"` even for a 6-channel output.

.github/workflows/ci.yml
```diff
@@ -117,7 +117,12 @@ jobs:
 
       - name: Run pip-audit for vulnerabilities
         run: |
-          uv run pip-audit --desc on --skip-editable --ignore-vuln PYSEC-2022-42969 --ignore-vuln CVE-2025-71176
+          uv run pip-audit --desc on --skip-editable \
+            --ignore-vuln PYSEC-2022-42969 \
+            --ignore-vuln CVE-2025-71176 \
+            --ignore-vuln CVE-2026-34993 \
+            --ignore-vuln CVE-2026-47265 \
+            --ignore-vuln PYSEC-2026-196
```
nghorbani: These three new ignores (CVE-2026-34993, CVE-2026-47265, PYSEC-2026-196) are unrelated to the selector change and are better split into a deps PR, matching how #22 handled CVEs. The two aiohttp advisories (34993 deserialization, 47265 cookie leak on redirect) are both fixed by `aiohttp >= 3.14.0`; PYSEC-2026-196 is a pip path traversal fixed by `pip >= 26.1.2`. Upgrading clears all three, which is preferable to suppressing them. If any must stay ignored, please add a one-line comment per entry stating the package, why it is deferred, and when to revisit.

cuvis_ai/node/channel_selector.py
```diff
@@ -137,8 +137,10 @@ class ChannelSelectorBase(Node):
     OUTPUT_SPECS = {
         "rgb_image": PortSpec(
             dtype=torch.float32,
-            shape=(-1, -1, -1, 3),
-            description="Composed RGB image [B, H, W, 3] in 0-1 range",
+            shape=(-1, -1, -1, -1),
```
nghorbani: Relaxing the shape here on the base class also loosens the contract for every other selector that inherits it (FastRGBSelector, RangeAverageFalseRGBSelector, the NDVI selectors), all of which always emit exactly 3 channels. Only FixedWavelengthSelector is variable-width, so consider overriding OUTPUT_SPECS on that subclass and leaving the base at `(-1, -1, -1, 3)` so the others keep validation.

cuvis_ai/node/channel_selector.py
```diff
@@ -606,10 +614,15 @@ class FixedWavelengthSelector(ChannelSelectorBase):
...
         self,
-        target_wavelengths: tuple[float, float, float] = (650.0, 550.0, 450.0),
+        target_wavelengths: tuple[float, ...] = (650.0, 550.0, 450.0),
         normalize_output: bool = True,
         **kwargs,
     ) -> None:
+        target_wavelengths = tuple(float(w) for w in target_wavelengths)
+        if len(target_wavelengths) < 1:
+            raise ValueError(
+                "FixedWavelengthSelector: target_wavelengths must contain at least one wavelength"
```
nghorbani, line 655: Good to validate here. Since `n != 3` is now reachable, this is also the right place to reconcile it with `norm_mode`: statistical and running modes assume 3 channels downstream (3-element running buffers, `reshape(-1, 3)` in `statistical_initialization`). Either reject those modes for `n != 3` here, or generalize the buffers/reshape to `n`. As written, `FixedWavelengthSelector(..., norm_mode="statistical")` with `n != 3` still reaches the unguarded fit path.

cuvis_ai/node/channel_selector.py
```diff
@@ -651,10 +664,25 @@ def forward(
...
+            if self.normalize_output and n != 3:
+                from loguru import logger
+
+                logger.warning(
+                    "FixedWavelengthSelector: normalize_output=True is only supported for "
+                    "3-channel selection; got {} target wavelengths. "
+                    "Returning raw stacked bands without normalization. "
+                    "Pass normalize_output=False to suppress this warning.",
+                    n,
+                )
```
nghorbani: Two things on this branch:
- The warning fires on every `forward` call and imports loguru inside the function. In a video run that is one identical warning per frame. Both `normalize_output` and `n` are known at construction, so emit this once in `__init__` and import loguru at module level (matching the other node modules).
- The comment correctly notes the `reshape(-1, 3)` / 3-element-buffer constraint, but the gate only covers `forward`. `statistical_initialization` (selected by `norm_mode="statistical"`, independent of `normalize_output`) still calls `reshape(-1, 3)` on the n-channel raw output: wrong stats for n=6, `RuntimeError` for n=4.
- Just below, `band_info["normalized_output"]` is set to `self.normalize_output`, so it reports `True` here even though the bands were returned raw. It should reflect `self.normalize_output and n == 3`.

rajanish4:
Thanks for the careful review @nghorbani — all 8 points addressed in `be653e2`. Conservative fixes throughout, no scope creep. Mapping:

### Most important

**1. Statistical/running fit path unguarded for n != 3 — fixed.**
Raise `ValueError` in `FixedWavelengthSelector.__init__` when `norm_mode in {STATISTICAL, RUNNING}` and `len(target_wavelengths) != 3`. Generalising the 3-element running buffers + `reshape(-1, 3)` would be a larger separate change; the conservative path here is to make the failure mode loud and obvious — `per_frame` is the only mode supported for n != 3 today. Documented in both the class docstring and the error message. Before pushing I audited every `norm_mode in {statistical, running}` call site across cuvis-ai-dinomaly + cuvis-ai-cookbook: all are `CIRSelector` / `CIETristimulusRGBSelector` (3-channel only) — the new guard breaks zero existing pipelines.

**2. `band_info["normalized_output"]` honesty — fixed.**
Now set to `self.normalize_output and n == 3`. Stored on `self._effective_normalize_output` at construction so both the forward branch and the `band_info` flag use the same value (no chance of divergence).

**3. OUTPUT_SPECS override moved to subclass only — fixed.**
`ChannelSelectorBase.OUTPUT_SPECS` restored to `shape=(-1,-1,-1,3)`. `FixedWavelengthSelector` now overrides with `(-1,-1,-1,-1)` via `{**ChannelSelectorBase.OUTPUT_SPECS, "rgb_image": …}` so it inherits the other ports. The 11 sibling fixed-3-channel selectors (`FastRGBSelector`, `RangeAverageFalseRGBSelector`, `_NormalizedDifferenceIndexBase` and its NDVI variants, `CIRSelector`, `CIETristimulusRGBSelector`, `HighContrastSelector`, `CameraEmulationFalseRGBSelector`, the 3 supervised selectors) keep their pipeline-validation safety net.

**4. Per-frame warning + in-function loguru import — fixed.**
Warning hoisted to `__init__` (both `n` and `normalize_output` are known there and don't change later — emitted exactly once). `from loguru import logger` moved to module-level alongside the other module imports. Verified by a new test that captures a loguru sink and asserts exactly 1 warning at `__init__` + 0 additional warnings across 5 forward calls.

**5. pip-audit ignores — reverted.**
The three new `--ignore-vuln` flags (`CVE-2026-34993`, `CVE-2026-47265`, `PYSEC-2026-196`) dropped from `.github/workflows/ci.yml`. You're right that they don't belong in this PR — they're about `aiohttp` and `pip`, not the selector. I'll open a separate deps PR upgrading `aiohttp >= 3.14.0` and ensuring `pip >= 26.1.2` rather than suppressing.

### Minor

**6. Docstring Ports section — updated** on both `ChannelSelectorBase` (now explains why the 3-channel contract is kept tight on the base) and `FixedWavelengthSelector` (now documents the `(-1,-1,-1,-1)` override and the n-channel band-stack behaviour).

**7. Tests for the n != 3 statistical/running path — added.**
- `TestNormModeGuard`: 8 parametrised cases — statistical/running raise at `__init__` for n ∈ {4, 6}; both modes remain allowed for n=3; `per_frame` allowed for n ∈ {1, 4, 6, 9}.
- `TestBandInfoHonesty`: 3 cases asserting `normalized_output` reflects reality + the strategy-label branch.
- `TestNormalizeOutputWarning::test_six_channel_normalize_true_warns_once_at_construction`: loguru-sink-based test verifying once-at-init semantics.
- `TestNormalizeOutputWarning::test_three_channel_normalize_false_does_not_warn`: regression for the 3-channel default not emitting the warning.
- `TestOutputSpecsShape` rewritten to assert the conservative split (base stays `(-1,-1,-1,3)`; subclass overrides to `(-1,-1,-1,-1)`).

Selector tests: 31/31 pass (up from 16 in the original PR). Full suite: 902/904 (2 pre-existing skips, no regressions).

**8. `band_info["strategy"]` — branches on n.**
`"baseline_false_rgb"` for n == 3 (preserved for BC with existing logs/consumers), `"stacked_bands"` for n != 3.

Diff: +271 / −23 across 3 files (`channel_selector.py`, the test file, `ci.yml`).

PR #33, opened 2026-05-15
feat(plugins): add augment plugin manifest (v0.1.1)

rajanish4:
## Summary

- Adds `configs/plugins/augment.yaml` registering the `cuvis-ai-augment` plugin at tag `v0.1.1`
- Plugin provides `AugmentationCompose` — a single Node that applies stochastic spatial augmentations (HFlip, VFlip, Rot90, SpatialCrop) to `(B, H, W, C)` hyperspectral cubes and paired masks, exclusively during `ExecutionStage.TRAIN`

## Manifest

```yaml
plugins:
  augment:
    repo: "https://github.com/cubert-hyperspectral/cuvis-ai-augment.git"
    tag: "v0.1.1"
    provides:
      - cuvis_ai_augment.node.compose.AugmentationCompose
```

## Verification

Verified live via `NodeRegistry.load_plugins()` with `tag: v0.1.1`:

```
Loaded plugins: ['augment']
```

Plugin installs cleanly from the tagged release, all dependencies resolve, and `AugmentationCompose` imports without error.

## Links

- Plugin repo: https://github.com/cubert-hyperspectral/cuvis-ai-augment
- Release v0.1.1: https://github.com/cubert-hyperspectral/cuvis-ai-augment/releases/tag/v0.1.1
- Tutorial notebook: `notebooks/use_cases/lentils_augmentation.ipynb`

PR #46, opened 2026-06-24
feat(plugins): refresh manifest pins, register augment, add pin auto-bump CI

nghorbani:
## Summary

Refreshes the plugin manifest pins to the latest releases, registers the augment plugin, and adds CI that keeps the pins current automatically. Targets `0.11.0`.

## Manifest pin refresh (`configs/plugins/`)

- adaclip `v0.1.4` -> `v0.1.5`
- dinomaly `v0.1.5` -> `v0.2.0`
- deepeiou `v0.1.2` -> `v0.2.0`
- trackeval `v0.1.2` -> `v0.1.3`
- ultralytics `v0.1.2` -> `v0.1.3`
- cuvis_ai_dataloader `v0.1.0` -> `v0.2.0`

These are tag-only bumps: the releases carried no node or port changes, so each manifest's `capabilities:` block is unchanged.

## sam3: local path -> published tag

`configs/plugins/sam3.yaml` swaps from the local dev `path:` to `repo:` + `tag: v0.1.7`. v0.1.7 relaxes `setuptools<83`, resolving the `setuptools>=81` conflict that had forced the local checkout. No released-plugin manifest now carries an uncommitted local path.

## augment registration

New `configs/plugins/augment.yaml` (capabilities format, `tag: v0.3.0`) exposing `AugmentationCompose` for training-time cube/mask augmentation, plus `tests/plugins/test_augment_manifest_sync.py`. This **supersedes #33**, which added an augment manifest in the deprecated nested `plugins:`/`provides:` format pinned to v0.1.1.

## Auto-bump CI

- `scripts/bump_plugin_pins.py` + `.github/workflows/plugin_pin_bump.yml` open a PR whenever a pinned plugin publishes a newer release (cron + manual dispatch). It authenticates as a GitHub App (`PIN_BUMP_APP_ID` / `PIN_BUMP_APP_PRIVATE_KEY`) so the bot's PR triggers CI; the job is guarded on the App id being present. The App is provisioned and scoped to this repo only; the workflow activates once this lands on `main`.
- The per-plugin manifest-sync tests now assert the pinned tag's *shape* (`vX.Y.Z`) instead of a frozen value, so a refresh touches only YAML.
- Fixed `scripts/fetch_plugin_pyprojects.py` to read the one-file manifest format; it had silently skipped every manifest in the registry-compat audit.

## Verification

- `ruff check` + `ruff format --check`: clean
- All plugin manifests load via `cuvis_ai_schemas.plugin.load_plugin_manifest`
- `pytest -m "not slow and not gpu"`: 944 passed, 2 skipped
- `bump_plugin_pins.py` dry-run: no pending bumps (pins are current)

### DIFF
--- a/.github/workflows/no-local-sources.yml
+++ b/.github/workflows/no-local-sources.yml
@@ -1,0 +1,40 @@
+name: No Local Sources
+
+# Fails CI if pyproject.toml declares a local [tool.uv.sources] path entry. A machine-specific
+# editable path must never ship in a release: it breaks a fresh `uv sync` / install of the tag.
+# Commented-out path entries are fine -- tomllib never sees them.
+
+on:
+  pull_request:
+  push:
+    branches: [main, master]
+
+permissions:
+  contents: read
+
+jobs:
+  no-local-sources:
+    name: Reject local uv.sources paths
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Check pyproject.toml
+        run: |
+          python - <<'PY'
+          import pathlib, tomllib
+          pp = pathlib.Path("pyproject.toml")
+          if not pp.exists():
+              print("no pyproject.toml; nothing to check")
+              raise SystemExit(0)
+          sources = tomllib.load(pp.open("rb")).get("tool", {}).get("uv", {}).get("sources", {})
+          local = {k: v for k, v in sources.items() if isinstance(v, dict) and "path" in v}
+          if local:
+              print("ERROR: local [tool.uv.sources] path entries must not be committed:")
+              for name, spec in local.items():
+                  print(f"  {name} = {spec}")
+              raise SystemExit(1)
+          print("OK: no local path sources in [tool.uv.sources]")
+          PY
--- a/.github/workflows/plugin_pin_bump.yml
+++ b/.github/workflows/plugin_pin_bump.yml
@@ -1,7 +1,8 @@
 name: Plugin Pin Bump
 
 # Opens (or updates) a PR that bumps each configs/plugins/*.yaml tag pin to the plugin's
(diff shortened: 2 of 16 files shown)


rajanish4 (commented):
Reviewed this — clean overall and CI is green. One change I'd request, plus one out-of-scope heads-up.

### Please pin augment to `v0.3.1` instead of `v0.3.0`

`configs/plugins/augment.yaml` pins `tag: v0.3.0`, but that released tag still carries a machine-specific source in its `pyproject.toml`:

```toml
[tool.uv.sources]
cuvis-ai-core = { path = "../../cuvis-ai-core/cuvis-ai-core", editable = true }
```

augment `v0.3.1` exists specifically to fix exactly this — from its release notes:

> Commented out the local `[tool.uv.sources]` cuvis-ai-core editable path so the released tag no longer carries a machine-specific source… Added a release gate that rejects any active local `[tool.uv.sources]` path entry.

The node set is identical across v0.3.0 and v0.3.1 (`cuvis_ai_augment.node.compose.AugmentationCompose` only), so the bump is free and `test_augment_manifest_sync.py` passes unchanged. It's also the same "no machine-specific path in a released tag" goal as the sam3 change in this PR, so worth being consistent.

Heads-up that the manifest audit on this PR doesn't catch it: `registry_compat` only checks dependency specifiers (`cuvis-ai-core>=0.10.0` satisfies the lock fine) and never `uv sync`s the plugin, so the `[tool.uv.sources]` path slips through; the runtime smoke test that would exercise provisioning is weekly-only.

### Out of scope, just flagging

`configs/plugins/detr.yaml` and `configs/plugins/turbovec.yaml` carry hard-coded absolute Windows dev paths (`path: "D:/code-repos/..."`) that only resolve on the author's machine. Not introduced by this PR and they're unreleased WIP plugins, so nothing to do here — just worth a separate cleanup at some point.

Otherwise LGTM.

rajanish4 (approved):

PR #52, opened 2026-07-08
feat: extend the node catalog and make the docs catalog plugin-aware (0.11.0)

nghorbani:
## Summary

Extends the built-in node catalog with ~30 new pure-torch nodes across seven families, adds a metal-scrap classification cookbook notebook, and makes the docs node catalog plugin-aware. The `CHANGELOG` is stamped `0.11.0`; after merge, tagging `v0.11.0` cuts the release.

## New node families

- **Spectral pretreatments** (`cuvis_ai/node/pretreatments/`): `SavitzkyGolay`, `ContinuumRemoval`, `SpectralDerivative`, `SNVCorrection`, `Logarithm`, plus fit-required `MeanCenter` / `UnitVarianceScaling` (streaming Welford). All composable `cube -> cube`.
- **Vegetation indices**: 11 selectors (`EVI`, `EVI2`, `SAVI`, `MSAVI`, `NDWI`, `NBR`, `GNDVI`, `NDRE`, `CIRedEdge`, `MCARI`, `PRI`) reusing the `NDVISelector` colormap machinery.
- **Unmixing**: `NNLSUnmixing` (validated against `scipy.optimize.nnls`) and blind `NMFUnmixing`.
- **Clustering**: `KMeansClusterer` / `GaussianMixtureClusterer` — sklearn fit frozen into torch buffers, pure-torch forward, optional foreground-`mask` fit.
- **Anomaly / thresholding**: `OneClassSVMDetector` (chunked pure-torch RBF decision function), `SaturatedPixelDetector`, `MultiRangeSlicer`, `IntensityThresholdSegmenter`.
- **Object-level inspection**: `BlobDetector`, `ShapeMorphology`, `SignaturesToReferences`, `MajorityVoteByBlob`, built on a shared OpenCV connected-components helper factored out of `MaskRobustifier`.
- **Patch inference & assembly**: `PatchSampler` + `ClassMapAccumulator` (dense per-pixel classification via a `frame_id`/`y`/`x` provenance contract), `ImageConcatenator`, `PngWriter`, and a per-frame `caption` port on `TitleOverlay`.

## Docs: plugin-aware node catalog

The published catalog showed "0 from plugins": the generator read `docs/data/plugin_sources.yaml`, a file that never existed, and failed silently. It now reads the plugin manifests in `configs/plugins/` directly (no plugin installs in the docs venv):

- plugin nodes render manifest-driven input/output port tables and a repo link pinned to the release tag,
- `kind: data_module` capabilities get their own rows and pill,
- the filter gains a Source facet (built-in / plugin / data module),
- an unparseable manifest or an empty plugin collection fails the docs build instead of shipping an empty list,
- the generator resolves both `configs/plugins/` and the packaged `cuvis_ai/configs/plugins/` layout, so it survives the config-packaging move.

## Example

`notebooks/use_cases/metal_scrap_classification.ipynb` rebuilds the Gursch et al. 2026 SWIR steel-scrap classifier as a Cuvis.AI pipeline (per-band standardizer, 3D-CNN, weighted cross-entropy, inverse-frequency class weights from the statistical phase, dense per-pixel inference, mask cleanup).

## Testing

- `uv run pytest -m "not slow and not gpu"`: 1125 passed (includes new per-node tests and `tests/docs/test_node_catalog.py` pinning the manifest-driven catalog collection).
- `uv run mkdocs build --strict` green; catalog renders 154 built-in + 50 plugin nodes + 11 data modules on this branch.
- `ruff format` / `ruff check` / `interrogate cuvis_ai/` clean via the pre-push hook.


cuvis_ai/node/clustering/gmm.py
```diff
@@ -0,0 +1,227 @@
...
+        covariance_type: str = "full",
+        reg_covar: float = 1e-6,
+        max_iter: int = 100,
+        n_init: int = 1,
+        random_state: int = 0,
+        **kwargs: Any,
+    ) -> None:
+        """Store mixture hyperparameters and register the fitted-state buffers."""
+        self.n_components = int(n_components)
+        self.covariance_type = str(covariance_type)
```
rajanish4, line 120: only full covariance works in forward, but we still accept diag/tied/spherical here. those give a different shaped precisions_cholesky_ and then forward() throws an IndexError, and the node is already marked initialized so a checkpoint saved in between is broken. can we reject != "full" in __init__ so it fails early? and add a pytest.raises for it.

cuvis_ai/node/pretreatments/savitzky_golay.py
```diff
@@ -0,0 +1,137 @@
...
+    _category = NodeCategory.TRANSFORM
+    _tags = frozenset({NodeTag.HYPERSPECTRAL, NodeTag.PREPROCESSING, NodeTag.TORCH})
+
+    INPUT_SPECS = {
+        "cube": PortSpec(
+            dtype=torch.float32,
+            shape=(-1, -1, -1, -1),
+            description="Input hyperspectral cube [B, H, W, C]",
+        ),
+        "wavelengths": PortSpec(
```
rajanish4, line 82: wavelengths is declared as a required input but forward never reads it. so for deriv>0 with non uniform band spacing the derivative magnitude is wrong (delta stays 1). SpectralDerivative does this correctly. either use the port or make it optional and document that delta = band spacing in nm. also would check window_length is odd.

cuvis_ai/node/pretreatments/logarithm.py
```diff
@@ -0,0 +1,70 @@
...
+
+        Returns
+        -------
+        dict[str, torch.Tensor]
+            ``{"cube": log_transformed}`` with the same shape as the input.
+        """
+        clamped = cube.clamp_min(self.eps)
+        if self.mode == "ln":
+            return {"cube": torch.log(clamped)}
+        return {"cube": torch.log10(clamped)}
```
rajanish4: docstring says reflectance -> (pseudo)absorbance but absorbance is -log10(R), this returns +log10 so the sign is opposite. add a negate option or reword the docstring. also mode is not validated (unknown value silently falls through to log10).

cuvis_ai/node/channel_selector.py
```diff
@@ -588,6 +588,788 @@ def forward(
...
+        """Resolve nearest spectral indices for every named band."""
+        wavelengths_np = np.asarray(wavelengths, dtype=np.float32)
+        if wavelengths_np.ndim == 2:
+            wavelengths_np = wavelengths_np[0]
+        if wavelengths_np.ndim != 1:
+            raise ValueError(f"Expected 1D wavelengths [C], got shape {wavelengths_np.shape}")
+        wavelengths_np = wavelengths_np.ravel()
+
+        indices = {
+            name: self._nearest_band_index(wavelengths_np, nm) for name, nm in self.band_nm.items()
```
rajanish4, line 995: this snaps to the nearest band with no tolerance, so an index needing SWIR (NBR, or a Gao-NDWI) on a VNIR only camera silently picks the top band and returns wrong values. can we warn if the nearest band is farther than some tolerance. also the additive-constant indices (EVI/SAVI/...) only make sense on reflectance, should be documented.

cuvis_ai/node/unmixing/nnls.py
```diff
@@ -0,0 +1,153 @@
...
+        "class_mask": PortSpec(
+            dtype=torch.int32,
+            shape=(-1, -1, -1),
+            description="1-based argmax abundance per pixel, 0 below min_total [B, H, W]",
+        ),
+    }
+
+    def __init__(
+        self,
+        max_iter: int = 200,
```
rajanish4: two things: no sum-to-one constraint (ANC only, not FCLS) - ok but please document it since people expect abundances summing to 1. and default max_iter=200 seems too low for close/collinear endmembers, the NMF path uses 2000.

cuvis_ai/node/unmixing/nmf.py
```diff
@@ -0,0 +1,178 @@
...
+        ),
+        "endmembers": PortSpec(
+            dtype=torch.float32,
+            shape=(-1, -1),
+            description="Learned endmember spectra [K, C]",
+        ),
+        "scores": PortSpec(
+            dtype=torch.float32,
+            shape=(-1, -1, -1, 1),
+            description="Per-pixel reconstruction residual ||x - W H|| [B, H, W, 1]",
```
rajanish4: this description ||x - W H|| doesnt match what the code computes - its the per-pixel reconstruction residual against the frozen endmembers, not that. small but confusing.

cuvis_ai/node/patch_inference.py
```diff
@@ -0,0 +1,316 @@
+"""Dense per-pixel classification: tile a cube into patches, then stitch predictions back.
+
+Two nodes form an inverse pair for patch-based per-pixel classification. ``PatchSampler`` turns a
```
rajanish4: these two are called an inverse pair but PatchSampler doesnt output frame_id/y/x/height/width, the accumulator gets those from a separate tiler datamodule. so they dont wire together directly (the notebook uses the plugin version). can we make the docstring/changelog clearer. also the accumulator keeps every frame map until reset() so memory grows on long streams.

cuvis_ai/node/deciders/multi_range_decider.py
```diff
@@ -0,0 +1,96 @@
...
+        ),
+    }
+
+    def __init__(
+        self,
+        edges: list[float] | None = None,
+        right: bool = False,
+        **kwargs: Any,
+    ) -> None:
+        self.edges = [float(e) for e in (edges if edges is not None else [0.25, 0.5, 0.75])]
```
rajanish4, line 76: no check that edges are increasing - torch.bucketize is silently wrong on unsorted boundaries. a small assert would help.

cuvis_ai/node/segmentation/intensity.py
```diff
@@ -0,0 +1,113 @@
...
+        reduction: str = "mean",
+        band_index: int = 0,
+        **kwargs: Any,
+    ) -> None:
+        if reduction not in self._VALID_REDUCTIONS:
+            raise ValueError(
+                f"reduction must be one of {self._VALID_REDUCTIONS}, got {reduction!r}"
+            )
+        self.low = float(low)
+        self.high = float(high)
```
rajanish4, line 82: no check that low <= high, an inverted interval just gives an all-zero mask silently.

rajanish4 (commented):
went through all the new nodes. i have 2 things we should fix before merge, the rest is smaller docs/robustness stuff which i left as inline comments.

**blockers**

1. `GaussianMixtureClusterer` only works with `covariance_type="full"` but it still accepts diag/tied/spherical at construction and fit. sklearn then returns a different shaped `precisions_cholesky_`, the node stores it and marks itself initialized, then forward() crashes with a cryptic IndexError - and a checkpoint saved in between has garbage buffers. can we reject != "full" in `__init__` so it fails fast? (and a pytest.raises for it)

2. the metal scrap notebook doesnt run for anyone except the original machine - dataset and pretrained weights are hardcoded absolute paths and with `RUN_TRAINING=False` it asserts the .pt exists. it also needs the external plugin at a fixed tag. the other use case notebooks got colab + huggingface download in 0.10.1, this one should follow the same so people can actually run it.

**smaller (detail in the inline comments)**
- SavitzkyGolay declares a `wavelengths` input it never uses -> derivative wrong on non uniform spacing
- Logarithm docstring says absorbance but returns +log10 (sign flipped)
- vegetation indices: reflectance-only assumption not documented, and nearest-band snapping has no tolerance (NBR on a VNIR camera is silently wrong)
- unmixing: no sum-to-one constraint documented, NNLS max_iter=200 too low, NMF scores docstring doesnt match the code
- PatchSampler/ClassMapAccumulator called an inverse pair but dont actually wire together, and the accumulator grows memory on long streams
- MultiRangeSlicer/IntensityThresholdSegmenter missing simple input validation, connected_components docstring mismatch, and SpectralSignatureExtractor is not actually new
- docs generator fix is good and it fails loud now, but the test pins no counts and has no failure-case test (which is the exact bug this PR fixes)

thanks, nice set of nodes overall.

cuvis_ai/utils/connected_components.py
```diff
@@ -0,0 +1,65 @@
+"""Shared connected-components labeling helper.
+
+cuvis.ai has no native torch connected-component labeling (CCL) op, so nodes
+that need instance labels round-trip a single 2-D frame through OpenCV's
+``cv2.connectedComponentsWithStats``.  Centralizing that here keeps the
```
rajanish4: docstring says connectedComponentsWithStats but the code actually calls connectedComponents.

tests/docs/test_node_catalog.py
```diff
@@ -0,0 +1,100 @@
+"""Tests for the docs node-catalog generator (``scripts/generate_node_catalog.py``).
```
rajanish4, line 1: this only checks things exist, no count is pinned - so a silent drop back to a few plugin nodes would still pass. and there is no test that a broken/empty manifest actually raises, which is the exact bug this PR fixes. would add both.

nghorbani:
Thanks for the review. Both blockers and all inline items are addressed and pushed (d28cd56 nodes, 184fc39 docs test, 5e59075 notebook).

**Blockers**

1. `GaussianMixtureClusterer.__init__` now raises `ValueError` for any `covariance_type != "full"` (with a `pytest.raises`), so it fails fast instead of storing a wrong-shaped `precisions_cholesky_` and crashing later in `forward`.
2. The metal-scrap notebook is reproducible: the hardcoded absolute dataset/weights/cache paths and the machine-specific assert are gone, a Colab badge + bootstrap cell were added, `DATASET_ROOT` resolves a Zenodo-provisioned folder (Colab-aware) with a soft "download it here" message, training is GPU-gated by default, and inference skips gracefully when no weights are available. Per our discussion the dataset comes from Zenodo (DOI 10.5281/zenodo.17076238) and the weights are not published, so the reproducible path trains on a GPU runtime (or you can point `PRETRAINED` at a local checkpoint).

On the smaller note: `SpectralSignatureExtractor` is a pre-existing node (added in v0.5.0); this PR only re-exports it from `cuvis_ai.node`, which is how the changelog already frames it. I added an import test to lock that re-export. Per-thread replies below.

cuvis_ai/node/clustering/gmm.py
nghorbani: Done in d28cd56. `__init__` now raises `ValueError` for any `covariance_type != "full"`, before any fit or buffer write, and a parametrized `pytest.raises` covers diag/tied/spherical.

cuvis_ai/node/pretreatments/savitzky_golay.py
nghorbani: Done in d28cd56. `wavelengths` is now an optional port; when connected, the derivative is rescaled to the median band step taken from it (the `delta` param is the fallback), and a warning fires on non-uniform spacing since a fixed savgol kernel cannot honor it exactly (SpectralDerivative stays the coordinate-aware option). Added an odd-`window_length` check and `mode` validation, plus tests for wavelength-driven spacing, the non-uniform warning, and both validations.

cuvis_ai/node/pretreatments/logarithm.py
nghorbani: Done in d28cd56. Added a `negate` option so `-log10(R)` (true absorbance) is reachable, fixed the module docstring sign claim, and `mode` now raises on unknown values instead of silently falling through to log10. Tests added for `negate` and the invalid-mode raise.

cuvis_ai/node/channel_selector.py
nghorbani: Done in d28cd56. Added a `band_tolerance_nm` param (default 50 nm); when the nearest band is farther than the tolerance the selector logs a warning once per band, so NBR's 2200 nm on a VNIR grid no longer resolves silently. Documented the reflectance-only assumption on EVI/EVI2/SAVI/MSAVI. Added a test for the NBR-on-VNIR warning and one asserting an exact grid stays quiet.

cuvis_ai/node/unmixing/nnls.py
nghorbani: Done in d28cd56. Documented in the module and class docstrings that this is ANC-only (non-negativity, no sum-to-one, not FCLS), and raised the default `max_iter` 200 -> 500 with a note that close/collinear endmembers may need more.

cuvis_ai/node/unmixing/nmf.py
nghorbani: Done in d28cd56. The `scores` output docstring now reads "per-pixel reconstruction residual ||E.T a - x|| against the frozen endmembers E", matching what the code computes.

cuvis_ai/node/patch_inference.py
nghorbani: Done in d28cd56. Dropped the "inverse pair" framing in the module and `ClassMapAccumulator` docstrings; they now state plainly that the `frame_id`/`y`/`x`/`height`/`width` keys come from a patch-tiler data module, not from `PatchSampler` (which emits only patches/labels, which is why the notebook uses the plugin's own versions). Also documented that the accumulator keeps one map per frame until `reset()`, so memory grows with the number of frames, and called out streaming eviction as out of scope. Changelog updated to match.

cuvis_ai/node/deciders/multi_range_decider.py
nghorbani: Done in d28cd56. `__init__` now raises `ValueError` when `edges` are not strictly increasing, before they reach `torch.bucketize`. Test added.

cuvis_ai/node/segmentation/intensity.py
nghorbani: Done in d28cd56. `__init__` now raises `ValueError` when `low > high` instead of silently returning an all-zero mask. Test added.

cuvis_ai/utils/connected_components.py
nghorbani: Done in d28cd56. Both docstring references now say `cv2.connectedComponents`, matching the actual call.

tests/docs/test_node_catalog.py
nghorbani: Done in 184fc39. Added a count-floor test (plugin nodes / data modules / built-in mirror) so a silent drop back to a handful fails, plus three failure-case tests: an empty manifest dir, an unparseable manifest, and manifests that collect zero capabilities all now assert a raise.

rajanish4 (approved):

PR #54, opened 2026-07-09
feat(metrics): pool average_precision over the epoch

rajanish4:
## What

`AnomalyDetectionMetrics.average_precision` accumulates across the epoch (reset only at the `(stage, epoch)` boundary), so each batch emits a *running* AP. Averaging that per-epoch — Lightning's default reduction — is batch-size-sensitive and wrong. Precision/recall/F1/IoU are genuinely per-batch and stay as-is.

## Change

- `POOLED_METRIC_NAMES = frozenset({"average_precision"})` — tells the trainer to skip it in per-batch logging.
- `compute_epoch_metrics()` — returns the exact pooled AP from accumulated torchmetrics state; the trainer logs it once at epoch end.

Purely additive and backward compatible: with an older trainer the node behaves exactly as before (the classvar and method are ignored).

## Test

`test_compute_epoch_metrics_returns_pooled_ap` — asserts AP is declared pooled, `compute_epoch_metrics()` is empty before any `logits` batch, and after two accumulated batches equals a single pooled `compute()`. `test_metrics.py`: 36 passed; node-module docstring coverage 100%.

## Depends on

The pooling takes effect end-to-end only with the trainer change in cuvis-ai-core#50 (epoch-end pooled logging). Mergeable independently; to activate it, the released cuvis-ai-core floor should include that change.

Refs: ALL-5850

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [Unreleased]
+
+- `AnomalyDetectionMetrics.average_precision` is now epoch-pooled through the trainer's native
+  metric-object logging: the node exposes the live `BinaryAveragePrecision` via `pooled_metrics()`
+  (replacing `compute_epoch_metrics()`), and the trainer logs it once per epoch with `on_epoch=True`,
+  so the reported AP is the exact pooled value rather than the per-batch running value. Requires a
+  `cuvis-ai-core` that includes the native `pooled_metrics()` logging.
+
 ## 0.10.1 - 2026-06-24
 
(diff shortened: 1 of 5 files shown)


nghorbani (approved):
Rebased onto main (picks up the #55 security lock) and floored to cuvis-ai-core>=0.10.1, the release carrying the native pooled_metrics() logging, with a clean relock (only core 0.10.0->0.10.1; torch/torchvision stay cu128; mistune/pillow/soupsieve stay on the patched versions). AnomalyDetectionMetrics.average_precision now pools correctly through the native trainer path. Good to squash-merge.

PR #57, opened 2026-07-14
cuvis-ai 0.11.1: core >=0.11.2 floor, plugin pins to latest releases, flat TrainingConfig, catalog + docs refresh

nghorbani:
Releases **cuvis-ai 0.11.1**, the catalog-repo half of the cuvis-ai-next train (schemas 0.8.0 / core 0.11.x already published).

- Raises the dependency floors to `cuvis-ai-core>=0.11.2` and `cuvis-ai-schemas[full]>=0.8.0` (folded `TrainingConfig`, current plugin/registry contracts).
- Repoints all ten plugin manifest pins to their latest tagged releases (adaclip v0.2.0, augment v0.3.3, dataloader v0.4.0, inspecscrap v0.2.2, deepeiou v0.2.1, dinomaly v0.4.1, rtsam2 v0.2.0, sam3 v0.2.1, trackeval v0.1.4, ultralytics v0.1.4), and adds the new AnomalyAUROCMetrics / AdaCLIPFocalDiceLoss capabilities.
- Flattens the trainrun/training configs onto the folded `TrainingConfig`; generalizes `LentilsAnomalyDataNode` into `AnomalyDataNode` (+ `class_mask` port, deprecated alias kept).
- Excludes unreleased local-path prototype manifests from the node catalog; packages `configs/` into the wheel so `CONFIG_ROOT` resolves on a pip install.
- Refreshes the plugin-development docs to the current single-file `name:` / `capabilities:` manifest shape and lists all ten official plugins.

Full notes in `CHANGELOG.md` (0.11.1).


cuvis_ai/configs/plugins/cuvis_ai_builtin.yaml
```diff
@@ -1168,6 +1168,53 @@ capabilities:
...
+      dtype: int32
+      shape: [-1]
+      optional: true
+      description: Wavelengths [C] in nm
+    mesu_index:
+      dtype: int64
+      shape: [-1]
+      optional: true
+      description: Measurement/frame index per batch element [B]
+  doc_summary: CU3S data node with binary anomaly label mapping.
```
birkholz-cubert, line 1227: **[BUG]** AnomalyDataNode manifest entry predates the class_mask port `[certain]`

This PR adds a `class_mask` input/output port to `AnomalyDataNode` (`cuvis_ai/node/data.py:129-136,144`) and changes the `mask` input's meaning to "Binary anomaly mask ... (0=normal, non-zero=anomaly)". The checked-in manifest entry for `cuvis_ai.node.data.AnomalyDataNode` here (and its `LentilsAnomalyDataNode` alias at line 1218) was never regenerated: `input_specs`/`output_specs` still stop at `mesu_index` with no `class_mask` key, and `mask`'s input description still reads "Multi-class segmentation mask [B, H, W]" (the pre-fix wording).

Per CLAUDE.md, plugin manifests are "the same files the pipeline loader and the gRPC server consume" for port specs — this is this repo's own `cuvis_ai_builtin` manifest, not an external plugin, so anything that trusts it (gRPC-composed pipeline validation/tooling, a host pipeline picker) for `AnomalyDataNode`'s I/O will not see `class_mask` at all, even though the live class supports it and it's unit-tested. The generated docs catalog is unaffected (it skips `cuvis_ai_builtin.yaml` entries in favor of live introspection per `scripts/generate_node_catalog.py:16`), but nothing else validates this manifest against the live class - `tests/plugins/test_plugin_contracts.py` only contract-checks `adaclip.yaml`.

CHANGELOG.md
```diff
@@ -2,9 +2,71 @@
...
+- **Added `PointPrompt`, an interactive point-prompt source node.** Emits a scheduled per-frame list
+  of `{element_id, x, y, type}` dicts (`type` in `positive` / `negative` / `neutral`) on a configured
+  `prompt_frame_id`, and an empty list on every other frame. Accepts `(x, y[, type])` tuples or
+  dicts, validates the type, and defaults to `positive`. Registered in
+  `configs/plugins/cuvis_ai_builtin.yaml`. Its output dict shape matches `SAM3PointExpansion`'s
+  `points` input, so it drives point expansion in a scripted (non-interactive) pipeline.
+- **Added the SAM3 single-frame point-expansion use case.** Two pipeline configs
+  (`configs/pipeline/sam3/sam3_point_expansion.yaml`, cu3s-sourced, and `…_video.yaml`, video-frame
+  sourced) plus the `notebooks/use_cases/object_selection_point_expansion.ipynb` walkthrough.
+- **Pinned the sam3 plugin to `v0.2.0`.** `configs/plugins/sam3.yaml` uses a `repo:` + `tag: v0.2.0`
```
rajanish4: **[docs]** this says `v0.2.0`, but the manifest ships `v0.2.1` `[certain]`

`cuvis_ai/configs/plugins/sam3.yaml` pins `tag: "v0.2.1"` (and the PR "Notes for review" also say v0.2.1) — this bullet is the only place that says v0.2.0, so the release notes misstate the shipped pin.

```suggestion
- **Pinned the sam3 plugin to `v0.2.1`.** `configs/plugins/sam3.yaml` uses a `repo:` + `tag: v0.2.1`
```

CHANGELOG.md
```diff
@@ -2,9 +2,71 @@
...
+  pin (restored from the temporary local `path:` checkout), the first tagged sam3 release to ship
+  `SAM3PointExpansion` and the `rgb_frame` -> `rgb_image` port rename.
+- **Registered the rtsam2 plugin.** Added `configs/plugins/rtsam2.yaml` pinned to
```
rajanish4, line 78: **[docs]** rtsam2 is described two contradictory ways in this section `[certain]`

This bullet says rtsam2 is "pinned to `v0.1.0`", but `configs/plugins/rtsam2.yaml` ships `tag: "v0.2.0"`; and the pin-bump bullet at L7 lists "rtsam2 v0.1.0 -> v0.2.0". rtsam2 has no manifest at the merge-base — it's brand new this release, so there was never a v0.1.0 to bump from. Suggest dropping rtsam2 from the "bumped pins" list (it's a new registration, not a bump) and stating the registration pin as v0.2.0 here.

scripts/validate_trainrun_configs.py
```diff
@@ -8,7 +8,7 @@
 import yaml
 from hydra import compose, initialize_config_dir
 
-CONFIG_ROOT = Path(__file__).resolve().parents[1] / "configs"
+CONFIG_ROOT = Path(__file__).resolve().parents[1] / "cuvis_ai" / "configs"
 TRAINRUN_DIR = CONFIG_ROOT / "trainrun"
 REQUIRED_DIRECTIVES = ("/pipeline@", "/data@", "/training@")
```
rajanish4: **[correctness]** `REQUIRED_DIRECTIVES` matches no shipped trainrun — the script reports FAIL for all of them `[certain]`

No trainrun uses a `/pipeline@` default (they declare `pipeline: ../pipeline/...` as a top-level key), and the statistical configs have no `/training@` default, so `python scripts/validate_trainrun_configs.py` exits 1 with "defaults missing composition entry containing '/pipeline@'" for essentially every file. Pre-existing (the PR only repointed `CONFIG_ROOT` on the line above) and *not* the CI gate — `tests/configs/test_trainrun_configs_valid.py` validates flatness directly and passes — but since the PR re-touches this file, worth fixing the directive list to the current trainrun shape or retiring the script in favour of the pytest.

scripts/bump_plugin_pins.py
```diff
@@ -43,7 +43,7 @@
 # and commented lines start with '#', so neither is matched.
 _TAG_LINE = re.compile(r'^(tag:[ \t]*)(["\']?)v\d+\.\d+\.\d+\2[ \t]*$', re.MULTILINE)
```
rajanish4: **[correctness]** this regex can't match sam3's tag line, so the auto-bumper silently skips sam3 `[certain]`

`_TAG_LINE` requires the line to end right after the (optionally quoted) semver (`[ 	]*$`), but `configs/plugins/sam3.yaml`'s pin line has a trailing comment: `tag: "v0.2.1"  # schemas 0.8.0 / core 0.11.0 floors ...`. So `subn` returns `count=0`, the script warns "found 0; skipped" and continues — it fails safe (no bad pin), but sam3 will never auto-bump; a future release needs a manual edit. Allowing an optional trailing comment (`...[ 	]*(?:#.*)?$`) fixes it.

cuvis_ai/node/video.py
```diff
@@ -448,6 +481,18 @@ def close(self) -> None:
...
+    def cleanup(self) -> None:
+        """Finalize the video file when the hosting pipeline is torn down.
+
+        A gRPC/session pipeline has no explicit driver ``close()`` call, so the
+        session-teardown ``cleanup()`` (invoked by ``CuvisPipeline.cleanup`` on
+        session close, pipeline replacement, or run stop) is where the ffmpeg
+        trailer gets flushed. ``close()`` is idempotent, so calling it here in
+        addition to an explicit driver ``close()`` is safe.
+        """
+        self.close()
```
rajanish4: **[correctness]** a failed trailer flush at session teardown is silently downgraded to a log warning `[certain]`

`cleanup()` calls `self.close()`, which raises `RuntimeError` on a non-zero ffmpeg return. But `cleanup()` is invoked by `CuvisPipeline.cleanup`, which wraps each `node.cleanup()` in try/except and only `logger.warning`s. So if ffmpeg fails to write the `moov` trailer during finalize in a gRPC/streaming session (no explicit driver `close()`), the pipeline logs a warning and disposes — leaving an unplayable mp4 while the run reports success. That no-explicit-close path is exactly what `cleanup()` was added to serve. Consider recording/surfacing a finalize failure rather than relying on the swallowed exception.

tests/docs/test_node_catalog_e2e.py
```diff
@@ -0,0 +1,242 @@
...
+    chip.click()
+    expect(catalog.locator("#node-filter-categories")).to_be_visible()  # stays open
+    expect(chip).to_have_attribute("aria-pressed", "true")
+    assert _count(catalog) < total
+    chip.click()
+    expect(chip).to_have_attribute("aria-pressed", "false")
+    assert _count(catalog) == total
+
+
+def test_two_tags_widen_results_or_semantics(catalog):
```
rajanish4: **[test-coverage]** OR-within is asserted only as `>=`, and AND-across-facets has no test at all `[certain]`

The filter predicate is correct (OR within a facet, AND across facets), but this test asserts `two_tags >= one_tag` — an accidental AND-within-facet regression that returned *equal* counts (or a dead second chip) would still pass. And no test selects a category **and** a tag to assert the result is the intersection (narrower than either alone). A semantics flip (AND<->OR within/across) would sail through CI. Suggest a strict OR assertion (two-tag count `>=` each independent single-tag count, with `>0` growth on a known pair) plus one AND-across case.

cuvis_ai/node/prompts.py
```diff
@@ -773,9 +773,110 @@ def forward(
...
+    _category = NodeCategory.SOURCE
+    _tags = frozenset({NodeTag.KEYPOINTS, NodeTag.INFERENCE, NodeTag.METADATA})
+
+    _VALID_TYPES = frozenset({"positive", "negative", "neutral"})
+
+    INPUT_SPECS = {
+        "frame_id": PortSpec(dtype=torch.int64, shape=(1,), description="Source frame index [1]."),
+    }
+    OUTPUT_SPECS = {
+        "points": PortSpec(
```
rajanish4, line 796: **[question]** can you confirm this `points` output contract matches `SAM3PointExpansion`'s input? `[speculative]`

`SAM3PointExpansion` lives in the external sam3 plugin, so the key name `points` and the `{element_id, x, y, type}` dict shape couldn't be cross-checked here. The design is internally consistent and mirrors `BBoxPrompt`, but if the consumer expects e.g. `point_coords`/`point_labels` tensors or a different key, the pipeline fails port-binding at build. Also note no shipped pipeline in this PR actually wires `PointPrompt` (the `points` port is optional / host-injected everywhere), so this contract is currently unexercised in-repo — worth a check against the plugin's `INPUT_SPECS` before someone wires it.

rajanish4:
**[release-hygiene]** three plugin manifests ship uninstallable local dev paths into the published catalog `[certain]`

`cuvis_ai/configs/plugins/detr.yaml` (`path: "D:/code-repos/cuvis-ai-detr/detr-init"`), `turbovec.yaml` (`path: "D:/code-repos/cuvis-ai-turbovec"`), and `bytetrack.yaml` (`../../../../cuvis-ai-bytetrack/bytetrack-init`) have no `repo:`/`tag:` pin. `scripts/generate_node_catalog.py:193` globs `*.yaml` with no source-type filter, so the published Nodes catalog renders these three as uninstallable rows (no repo link), while `docs/reference/plugin-development/overview.md` documents only the 10 official plugins — 13 catalog sources vs 10 documented. These are content-identical renames from `configs/plugins/` (pre-existing, shipped in 0.11.0 too — not a regression here), but since this is the release gate and the same PR touches the generator, it's the moment to pin them to `repo:`+`tag:` or exclude local-path manifests from the catalog. The `D:/...` absolute paths resolve on no other machine.

cuvis_ai/configs/plugins/cuvis_ai_builtin.yaml
nghorbani: Fixed in 9e45cf0. Regenerated both the `AnomalyDataNode` entry and its `LentilsAnomalyDataNode` alias in `cuvis_ai_builtin.yaml`: added the `class_mask` input (generic tensor, so the module's `uint8` ids bind without a strict-dtype rejection) and output (`int32 [B, H, W, 1]`), and corrected the `mask` input description to the binary-anomaly wording.

To stop this drifting silently again, also added `tests/plugins/test_builtin_manifest_sync.py`, which asserts the builtin manifest's port set for these two nodes matches the live `INPUT_SPECS`/`OUTPUT_SPECS`.

CHANGELOG.md
nghorbani: Fixed in 9e45cf0, applied your suggestion. Synced the PR description too, which mirrored the same line.

CHANGELOG.md
nghorbani: Fixed in 9e45cf0. Dropped rtsam2 from the "bumped pins" bullet (it's a new registration this release, not a bump) and corrected the registration line to `v0.2.0`. PR description synced as well.

scripts/validate_trainrun_configs.py
nghorbani: Fixed in 9e45cf0. It turned out to be broader than the directive list: 5 of the 18 trainruns don't follow the composition convention (2 are fully inlined with no `# @package _global_` and no `defaults:`, and 3 composition-style ones declare no pipeline at all), so the header and defaults checks were failing them too.

Reworked the structural checks to accept both shapes: require the `_global_` header and `_self_` only when a `defaults:` block is present, require `data` via a `/data@` default or an inline `data:` key, and lean on the Hydra compose as the authoritative check. `python scripts/validate_trainrun_configs.py` now reports OK for all 18.

scripts/bump_plugin_pins.py
nghorbani: Fixed in 9e45cf0. One tweak on the suggested pattern: `subn` replaces the whole match, so a non-capturing `(?:#.*)?$` would bump the version but strip sam3's trailing comment. I captured the trailing run as group 3 and re-append it (`...\2([ \t]*(?:#.*)?)$` with `\g<3>` in the replacement), so the pin bumps and the comment survives. Verified it round-trips `tag: "v0.2.1"  # ...` (count=1, comment intact).

cuvis_ai/node/video.py
nghorbani: Fixed in 9e45cf0. `cleanup()` now catches the finalize `RuntimeError`, logs it at ERROR with the output path, and records it on `self._finalize_error` before re-raising, so a failed trailer flush is surfaced above the pipeline's swallowed warning and is detectable by callers/tests instead of leaving a silently-truncated file. Added `test_to_video_node_cleanup_surfaces_finalize_failure` to cover it.

tests/docs/test_node_catalog_e2e.py
nghorbani: Fixed in 9e45cf0. Replaced the weak `>=` check: `test_tags_combine_as_or_within_facet` now asserts the OR bounds (`max(a, b) <= a|b <= a + b`) for every pair among the first few tags, so an AND-within regression breaks the lower bound, and requires at least one pair to strictly grow the result set, which a dead/duplicate second chip can't. Added `test_category_and_tag_intersect_across_facets` asserting a category+tag result is `<= min(cat_only, tag_only)` and `< total`, catching an OR-across flip. Both pass under `--runslow`.

cuvis_ai/node/prompts.py
nghorbani: Confirmed. Checked against sam3 v0.2.1 (`cuvis_ai_sam3/node/sam3_point_expansion.py`): `SAM3PointExpansion.INPUT_SPECS["points"]` is `dtype=list, shape=(), optional=True`, a per-frame list of `{element_id, x, y, type}` dicts, which is byte-for-byte what `PointPrompt.OUTPUT_SPECS["points"]` emits.

You're right that no shipped pipeline wires `PointPrompt` (the `points` port is host-injected in the sam3 point-expansion presets), but the notebook `notebooks/use_cases/object_selection_point_expansion.ipynb` does drive it, so the contract is exercised end to end there. No code change needed.

nghorbani:
> Re: **[release-hygiene]** three plugin manifests ship uninstallable local dev paths

Addressed in 9e45cf0. Went with excluding local-path manifests from the catalog and kept the files for local dev. `collect_plugin_nodes` in `scripts/generate_node_catalog.py` now skips any non-git manifest that isn't the `cuvis_ai_builtin` self-mirror, so `detr` / `turbovec` / `bytetrack` no longer render as repo-less rows and the catalog matches the 10 documented plugins. The builtin mirror still flows through the existing `exclude_dotted` dedup. Added `test_local_path_prototypes_excluded_from_catalog` as a guard.

Pinning them to `repo:` + `tag:` wasn't an option since none of the three has a published release yet.

rajanish4 (approved):

PR #58, opened 2026-07-14
feat(losses): add DiceLoss and CrossEntropyLoss segmentation loss nodes

rajanish4:
Adds the two dense-segmentation losses the builtin library lacked (IoULoss is binary-only): **DiceLoss** and **CrossEntropyLoss**, both consuming BHWC per-pixel class logits + an integer class-index mask and emitting a scalar loss.

## Design
- Class count is inferred from the logits' last axis — no `num_classes` hparam. `K=1` runs the sigmoid/binary Dice formulation, `K>1` softmax over one-hot targets.
- `DiceLoss`: `weight` / `ignore_index` / `include_background` / `eps`. Excluding background from the per-class mean is the standard nnU-Net-style choice for heavily imbalanced segmentation.
- `CrossEntropyLoss`: `weight` / `class_weights` (registered buffer, cast to logits dtype) / `ignore_index` (-100, PyTorch default).
- Both subclass the existing `LossNode` (train/val/test only) and follow the `IoULoss` pattern for the SEGMENTATION tag; registered in `configs/plugins/cuvis_ai_builtin.yaml`.
- These are the first builtin losses with multiclass `[B,H,W,K]` logits; existing ones are binary `[B,H,W,1]`.

## Provenance
Battle-tested in the cuvis-ai-unet plugin on the lentils foreign-object segmentation task (fg-IoU 0.79 champion). The plugin keeps its copies until this ships, then drops them in its next minor (pipeline configs reference fully-qualified class paths, and NodeRegistry resolves duplicates last-wins by simple name, so coexistence is safe).

## Tests
`tests/training/test_losses.py`: reference equivalence (hand-computed soft Dice; `F.cross_entropy` on permuted logits), weight scaling, gradient flow, `include_background`, `ignore_index`, single-logit binary head, trailing target channel squeeze — 36 total in the file, all green. `ruff` + `ruff format` clean, `interrogate` 96.8% (≥95 gate).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- **Added `DiceLoss` and `CrossEntropyLoss` segmentation loss nodes.** Multiclass-capable soft Dice (with an `include_background` toggle) and cross-entropy over BHWC logits `[B,H,W,K]` with integer targets `[B,H,W]`, subclassing the existing `LossNode` base (train/val/test stages) and registered in `cuvis_ai/configs/plugins/cuvis_ai_builtin.yaml`. Ports the two losses out of the cuvis-ai-unet plugin so it can drop its local copies.
 - Bumped the `cuvis_ai_builtin` manifest pin v0.11.5 -> v0.12.0 so composed child environments install the released cuvis-ai matching the host.
 
 ## 0.12.0 - 2026-08-20
@@ -124,8 +125,6 @@
   `CocoTrackMaskWriter`.
 - **Dropped the abstract `SAM3TrackerInference` entry from `configs/plugins/sam3.yaml`.** Base
   classes are not instantiable from pipelines, so the manifest now lists only concrete nodes.
-
-- **Added `DiceLoss` and `CrossEntropyLoss` segmentation loss nodes.** Multiclass-capable soft Dice (with an `include_background` toggle) and cross-entropy over BHWC logits `[B,H,W,K]` with integer targets `[B,H,W]`, subclassing the existing `LossNode` base (train/val/test stages) and registered in `configs/plugins/cuvis_ai_builtin.yaml`. Ports the two losses out of the cuvis-ai-unet plugin so it can drop its local copies.
 
 ## 0.11.0 - 2026-07-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,6 +125,8 @@
 - **Dropped the abstract `SAM3TrackerInference` entry from `configs/plugins/sam3.yaml`.** Base
   classes are not instantiable from pipelines, so the manifest now lists only concrete nodes.
 
+- **Added `DiceLoss` and `CrossEntropyLoss` segmentation loss nodes.** Multiclass-capable soft Dice (with an `include_background` toggle) and cross-entropy over BHWC logits `[B,H,W,K]` with integer targets `[B,H,W]`, subclassing the existing `LossNode` base (train/val/test stages) and registered in `configs/plugins/cuvis_ai_builtin.yaml`. Ports the two losses out of the cuvis-ai-unet plugin so it can drop its local copies.
+
 ## 0.11.0 - 2026-07-14
 
 - **Docs: the node catalog now lists plugin capabilities.** The Catalogs â†’ Nodes generator reads
--- a/cuvis_ai/configs/plugins/cuvis_ai_builtin.yaml
+++ b/cuvis_ai/configs/plugins/cuvis_ai_builtin.yaml
@@ -1593,7 +1593,7 @@ capabilities:
       dtype: int32
       shape: [-1, -1, -1]
       optional: false
-      description: Integer class-index mask [B, H, W] (cast to int64 internally; trailing [.,1] ok)
+      description: Integer class-index mask [B, H, W]; cast to int64 internally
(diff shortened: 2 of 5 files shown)


nghorbani (commented):
Anish, thanks for porting these over. The library needed proper dense-segmentation losses, and the test suite is exactly how I like these done (hand-computed soft-Dice reference, `F.cross_entropy` equivalence instead of snapshot numbers). The spec choices also line up with the codebase: `int32` rank-3 `[B,H,W]` targets match the batch `mask` convention, and both nodes follow the `IoULoss` pattern for base class, category and tags.

The branch had drifted 10 commits behind main (the plugin manifests moved to `cuvis_ai/configs/plugins/` in 0.11.1, hence the conflict flag), so I rebased it onto main and force-pushed, then put one fix commit (b7217be) on top addressing the points below. Please eyeball that commit; if you are fine with it, this is good to go from my side.

**P1, fixed in b7217be: `CrossEntropyLoss` silently returned zero loss for single-logit heads.** The docstring said `K >= 2` but nothing enforced it. Softmax over one logit is constant 1, so a `[B,H,W,1]` binary head (our standard binary shape, cf. `IoULoss` and the anomaly convention) yielded a loss that is identically 0 with zero gradients: training runs, logs step times, and learns nothing, with no error anywhere. `forward` now raises a `ValueError` pointing at `DiceLoss` / `AnomalyBCEWithLogits`, with a test.

**P2, fixed: Dice clamped out-of-range targets.** `_soft_dice_loss` did `safe.clamp(0, num_classes - 1)`, which silently folds labeling bugs (a stray class index 3 in a 3-class problem) into valid classes, while `CrossEntropyLoss` raises on the same input. I dropped the clamp so `F.one_hot` surfaces the bad label; `ignore_index` behaves as before. Also added the missing test for the binary (`K == 1`) Dice path with `ignore_index`.

**P2, no code change, flagging it: post-construction `ce.class_weights = [...]` is inert.** `forward` reads the `_class_weights` buffer; the attribute is only an hparam mirror. We hit exactly this trap in metalscrapes in June (loss ran unweighted, no error). Constructor and YAML round-trip are correct, so this is fine for the PR as is, but I would like a follow-up: an optional `class_weights` input port, so an inverse-frequency entry node can feed Phase-1-fitted weights through the graph (the pattern metalscrapes converged on). Opinion?

**P3, fixed: docs.** The targets spec description promised "trailing [.,1] ok", but the declared rank-3 spec is what graph wiring binds, so the squeeze only helps direct `forward()` calls; reworded in the specs and the manifest. Also documented that Dice here is batch Dice (accumulated per class over the whole batch, nnU-Net style) rather than a per-sample mean, since both formulations are common and callers should know which one they get.

Rebase notes: the changelog bullet got fuzzed into the released 0.11.1 block during the rebase; it now sits in a fresh `[Unreleased]` section at the top. Your manifest entries survived the `configs/` move byte-for-byte against a regen. While checking that, I noticed `emit_metadata --check` is already stale on main for unrelated nodes (icon palette, `ClassMapToRGB` int64, the caption port), so that needs a separate manifest refresh on main; it is not this PR's problem.

Everything green locally: 39 loss tests, ruff format and check, full pre-push gate.

rajanish4:
Reviewed `b7217be` — all three fixes look right, good to merge from my side:

- **P1 (CE single-logit → ValueError):** agreed, and it was a real silent-zero-gradient trap; pointing the error at `DiceLoss` / `AnomalyBCEWithLogits` is the right nudge.
- **P2 (drop the Dice clamp):** yes — silently folding an out-of-range label into a valid class is worse than a loud `one_hot` failure, and it now matches CE.
- **P3 (batch-Dice + spec wording):** good to state the batch-vs-per-sample distinction explicitly; the rank-3 spec is what graph wiring binds, so the squeeze caveat is accurate.

On the `class_weights` follow-up (inert post-construction attribute): agreed it's worth closing — running unweighted with no error is exactly the footgun. My lean is to fit `class_weights` as a **statistically-initialized buffer** (Phase-1 inverse-frequency over the label histogram, stored in the `state_dict`) rather than a runtime input port: the weights are fixed node state after fitting, so a buffer keeps them serialized + reproducible without threading a per-forward port that reads as dynamic per-batch weighting. If you'd rather match the metalscrapes entry-node + port pattern for consistency, that's fine too — I'd just constrain the port to a fixed fitted tensor set once at Phase-1, not recomputed per batch. Either way the immediate footgun deserves a guard: have the `class_weights` setter update `_class_weights` (or warn) so `ce.class_weights = [...]` isn't silently inert. Happy to take the follow-up.

nghorbani (approved):
Rebased onto main (post-0.12.0) and re-verified. The two conflicts resolved as expected: the changelog entry now sits under the current Unreleased section next to the pin-bump line, and the manifest keeps the git-tag header with both new capability entries at their alphabetical spots.

Local verification in a fresh worktree venv on main's lock: `emit_metadata --check` leaves both new entries byte-identical under regen (the remaining drift it reports is pre-existing on main and tracked separately), ruff format/check clean, interrogate 96.9%, 1295 tests green. All CI checks pass on the rebased head, including the security scan that was red on the stale base.

One note for the release train: the builtin manifest pins tag v0.12.0, which predates these classes, so composed child envs can only load DiceLoss/CrossEntropyLoss after the next release plus pin bump. Same window as any node added between releases.

PR #63, opened 2026-07-22
fix(normalization): accept non-contiguous inputs (view -> reshape)

rajanish4:
## Problem

A pipeline feeding a spatially cropped tensor (e.g. a lane-crop node) into `MinMaxNormalizer` crashes with:

```
RuntimeError: view size is not compatible with input tensor's size and stride
(at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead.
```

## Root cause

The normalizers in `cuvis_ai/node/normalization.py` flatten BHWC inputs with `Tensor.view(B, -1, C)`, which requires contiguous memory. Any cropped, sliced, or permuted upstream tensor is non-contiguous and breaks the flattening step.

## Fix

Replace the data-tensor `.view(...)` calls in `MinMaxNormalizer`, `SigmoidNormalizer`, and `PerPixelUnitNorm` (`_normalize` flatten + un-flatten) with `.reshape(...)`. `reshape` is semantically identical and only copies when the input is non-contiguous, so the contiguous fast path is unchanged. The stat-vector/buffer `.view(1, 1, 1, C)` calls in `PercentileNormalizer` operate on freshly created contiguous tensors and are left as-is.

## Test

`tests/node/test_normalization_noncontiguous.py` builds a non-contiguous BHWC crop (`torch.rand(1, 64, 64, 8)[:, 5:60, 3:50, :]`) and asserts `MinMaxNormalizer(use_running_stats=False)`, `SigmoidNormalizer`, and `ZScoreNormalizer` return finite, shape-preserving output. The new test plus all existing normalization tests pass (46 passed).

Refs ALL-5942


nghorbani (approved):
Anish, the fix is correct and I am approving. What I checked and what I pushed on top:

**Verified.** I audited every remaining `.view` call in `normalization.py`. The six data-tensor flatten/un-flatten calls you replaced were exactly the unsafe ones; the `PercentileNormalizer` stat-vector `.view(1, 1, 1, C)` calls operate on freshly created contiguous tensors, so leaving them was the right call. `reshape` is a no-op alias for `view` on contiguous input, so the hot path is unchanged.

One note on the test matrix: `ZScoreNormalizer` never flattens (pure dim-reductions), so it was already safe with non-contiguous input. Keeping it as a regression guard is fine, but `PerPixelUnitNorm` was changed by this PR and not covered, so I added it to the parametrize list.

**Pushed to your branch** (brought up to date with main first):
- `PerPixelUnitNorm` added to `test_normalizers_accept_non_contiguous_input`.
- CHANGELOG entries under a new `## Unreleased` section.
- The Security Scanning fix: `gitpython` floor raised to >=3.1.55 (locked 3.1.59), plus refreshed `jupyterlab` (4.6.3) and `aiohttp` (3.14.3) locks. That job was failing on freshly published advisories against the old lock, which predates your diff entirely, but it was blocking a green merge here so I bundled the bumps. Lock diff is limited to those three packages; the cu128 torch pins are untouched, and pip-audit with the CI flags is clean locally.

Will squash-merge once all checks are green.

PR #67, opened 2026-08-13
fix(configs): source cuvis_ai_builtin from git, not a deployment-dependent path

birkholz-cubert:
## What

Four changes, three commits:

1. **`cuvis_ai/configs/plugins/cuvis_ai_builtin.yaml`** - the fix. `path: "../../.."` is replaced by a git tag pin:

   ```yaml
   name: cuvis_ai_builtin
   package_name: cuvis-ai
   repo: https://github.com/cubert-hyperspectral/cuvis-ai.git
   tag: "v0.11.5"
   ```

   `package_name` is required: for a git source `resolve_plugin_sources` otherwise falls back to the manifest key, uv normalises that to `cuvis-ai-builtin`, and the install dies on uv's metadata name check.

2. **`tests/plugins/test_builtin_manifest_location_independent.py`** (new) - copies the manifest to another directory and requires the parsed source to be identical. Tests the invariant, not the specific source kind.

3. **`cuvis_ai/configs/pipeline/anomaly/dinomaly/dinomaly_cir.yaml`** - rewritten as an inference-only preset (`metadata.name: dinomaly_cir_lentils`) matching the published lentils checkpoint: two-stage decider (`image_threshold: 0.14`, `top_k_fraction: 0.001`), `running_warmup_frames: 0`, and the training-only nodes removed (loss bridge, both metric nodes, heatmap visualizer, TensorBoard monitor).

4. **`pyproject.toml` + `uv.lock`** - `[tool.uv.sources]` temporarily points `cuvis-ai-core` at the `fix/child-runtime-cuda-torch` branch (cuvis-ai-core PR #59), locked to `0.12.0.post2+g6ac3bf3a6`. Marked LOCAL ONLY in the file.

Items 3 and 4 are unrelated to the manifest fix and are flagged below.

## Why

`path: "../../.."` resolves against the manifest's own directory, so it means two different things:

| deployment | resolves to | buildable |
| --- | --- | --- |
| source checkout | `<checkout>` | yes |
| wheel install | `<site-packages>` | **no** |

The configs ship inside the wheel (`package-data: configs/**/*.yaml`), so a customer install always takes the second row, and it fails outright rather than degrading:

```
RuntimeProjectError: Local plugin 'cuvis_ai_builtin' at <site-packages> has no
pyproject.toml; the composer needs it to learn the package name.
```

A tag pin needs no path resolution and therefore cannot vary by deployment.

Rejected alternative: repoint the path at the package directory and ship `pyproject.toml` as package data. `version` is dynamic via setuptools-scm and needs a git repo, `packages.find` is relative to the project root so a build from `<site-packages>/cuvis_ai` would not contain `cuvis_ai`, and it needs a second doctored pyproject committed inside the package.

## Verified

- **Pre-fix repro** against a real non-editable install of this checkout: the shipped manifest resolved to `<site-packages>` and raised the error above.
- **Post-fix, both deployments** through the real loader: wheel install and dev checkout both yield `GitPluginSource(package_name=cuvis-ai, tag=v0.11.5, sha=d127e03f0910)`.
- **Real `compose_env`** with the wheel-installed manifest and a registry core (`cuvis-ai-core==0.12.0`, the clean-system shape): `uv.lock` gets `cuvis-ai 0.11.5` from `git ...?rev=d127e03f`, and the child interpreter imports `cuvis_ai 0.11.5` with `torch 2.11.0+cu128` and CUDA available. setuptools-scm resolves a clean `0.11.5` out of uv's git checkout, so the shallow-clone concern does not materialise.
- **`package_name` necessity** reproduced directly: without it uv reports `Package metadata name 'cuvis-ai' does not match given name 'cuvis-ai-builtin'`.
- **Tests**: `tests/plugins` + `tests/configs` = 41 passed, 13 skipped. The new regression test was confirmed to fail against the pre-fix manifest, with exactly the checkout-root vs site-packages divergence.

## Possible pitfalls

1. **Two commits in here must not merge as they stand.** The `cuvis-ai-core` git-branch source in `pyproject.toml` and the matching `uv.lock` entry are local development scaffolding for cuvis-ai-core PR #59. `[tool.uv.sources]` does not reach the built wheel's metadata, so a release would not ship the git URL - but CI syncs `--locked`, so main would fetch core from an unmerged branch, and the lock becomes unresolvable the moment that branch is deleted. Re-comment the line and relock before merge.
3. **Host/child version skew is now structural.** The pin is static, so a child can run a different cuvis-ai patch version than the host that composed it - and after this ships as 0.11.6 the child will still install v0.11.5, whose own floor is `cuvis-ai-core>=0.11.2`. The existing `plugin_pin_bump` workflow picks this manifest up now that it has `repo` + `tag`, so the pin tracks releases without manual edits. Exact parity needs a registry source kind in the manifest schema, which is a cross-repo change and not this PR.
4. **The child now needs git and reachable github.com** to build the builtin plugin. Other manifests are already git-sourced, so this is not a new class of requirement, but an offline or air-gapped install has no path to the builtin nodes at all.
5. **The child no longer gets `cuvis-ai` as an editable path member,** so it no longer inherits cuvis-ai's own `[tool.uv.sources]` cu128 index by accident. That accident is exactly what hid the CPU-torch bug on developer machines (ALL-6036, cuvis-ai-core PR #59). Removing the leak is only safe together with that change, which pins child torch explicitly.
6. **The regression test resolves its manifest relative to the CWD** (`Path("cuvis_ai/configs/plugins/...")`), so it passes under CI's repo-root invocation and fails from anywhere else. `Path(__file__).parents[2]` would make it location-independent, like the thing it tests.

## Out of scope, filed separately

Three manifests still carry local paths, two of them absolute developer paths: `detr.yaml` and `turbovec.yaml` point at `D:/code-repos/...`, `bytetrack.yaml` at a sibling checkout. They ship in the wheel and cannot resolve on any other machine. Same class of defect, tracked separately.

### DIFF
--- a/cuvis_ai/configs/data/lentils.yaml
+++ b/cuvis_ai/configs/data/lentils.yaml
@@ -4,8 +4,11 @@
 
 data_module: cu3s
 splits:
-  # Selectors over the cu3s measurement universe; val/test disjoint so the default
-  # leakage_check (error) passes.
+  constraints:
+  - kind: no_split_overlap
+    severity: error
+  # Selectors over the cu3s measurement universe; val/test disjoint so the declared
+  # no_split_overlap (error) constraint passes.
   train:
     - {kind: file_indices, source: data/Lentils/Lentils_000.cu3s, ids: [0, 2, 3]}
   val:
--- a/cuvis_ai/configs/data/tracking_cap_and_car.yaml
+++ b/cuvis_ai/configs/data/tracking_cap_and_car.yaml
@@ -1,5 +1,8 @@
 data_module: cu3s
 splits:
+  constraints:
+  - kind: no_split_overlap
+    severity: error
   # Frame IDs - fill after running inspect mode
   train:
     - {kind: file_indices, source: &src "D:\\data\\2024_05_22_cap_and_car_2XM_setup\\2024_05_22__cap_and_car_50mm\\Auto_002.cu3s", ids: [696, 708, 720, 855, 858, 871, 881, 887]}
--- a/cuvis_ai/configs/pipeline/anomaly/dinomaly/dinomaly_cir.yaml
+++ b/cuvis_ai/configs/pipeline/anomaly/dinomaly/dinomaly_cir.yaml
@@ -1,20 +1,17 @@
 metadata:
-  name: dinomaly_cir_lentils
-  description: "Inference-only variant of dinomaly_cir, wired to match the published
-    lentils checkpoint (XMR_Demo_Industrial_Foreign_Object_Detection_Lentils,
-    dinomaly_cir_full_pipeline/dinomaly_cir.pt). Same AnomalyDataNode -> MinMaxNormalizer
-    -> 860/670/560 selector -> DinomalyDetector spine as the training preset, but with
-    that checkpoint's two-stage decider calibration and without the training-only
-    metrics, heatmap and TensorBoard nodes. Load it with the .pt: the normalizer's
-    running stats live there, not in this file."
+  name: dinomaly_cir
+  description: "Dinomaly anomaly detection on a fixed-wavelength CIR (color-infrared)
+    false-color projection (860/670/560 nm = NIR/R/G): AnomalyDataNode → MinMaxNormalizer
+    → FixedWavelengthSelector → DinomalyDetector, with quantile decider, pixel/image
+    metrics, AUROC, and per-epoch score heatmaps logged to TensorBoard."
   created: ''
   tags:
+  - gradient
   - dinomaly
   - anomaly
   - cir
-  - inference
   author: cuvis.ai
-  cuvis_ai_version: 0.11.5
+  cuvis_ai_version: 0.11.1
 plugins:
 - dinomaly
 - cuvis_ai_builtin
@@ -28,10 +25,8 @@ nodes:
   class_name: cuvis_ai.node.normalization.MinMaxNormalizer
   hparams:
     eps: 1.0e-06
-    # Bounds come from the checkpoint (0 .. 18783, raw counts). Switching this to false
-    # would normalize per frame, which is not the distribution the detector was trained
-    # on and would invalidate the decider's image_threshold below.
     use_running_stats: true
+    max_initialization_frames: 20
 - name: cir_selector
   class_name: cuvis_ai.node.channel_selector.FixedWavelengthSelector
   hparams:
@@ -43,9 +38,7 @@ nodes:
     norm_mode: running
     apply_gamma: true
     freeze_running_bounds_after_frames: 20
-    # 0, not the preset's 10: the checkpoint carries per-channel bounds, so there is
-    # nothing to warm up and a warmup would ignore them for the first frames.
-    running_warmup_frames: 0
+    running_warmup_frames: 10
 - name: dinomaly_detector
   class_name: cuvis_ai_dinomaly.node.dinomaly_detector.DinomalyDetector
   hparams:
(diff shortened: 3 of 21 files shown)


nghorbani (commented):
Simon, the manifest fix is right and the verification section is exactly what I needed to see; the tag pin is self-consistent too (v0.11.5 already carries TwoStageBinaryDecider and its capability entry). I pushed the cleanup directly onto the branch (026a79d, rebased onto main) so this can merge. Here is what changed and why.

**The one thing the PR body did not flag: the dinomaly_cir rewrite breaks the CuvisNEXT training preset.** `configs/trainrun/dinomaly_cir_cuvisnext.yaml` resolves `loss_nodes: [dinomaly_train_loss]`, `metric_nodes: [metrics_anomaly, metrics_auroc]` and the checkpoint monitor `metrics_anomaly/iou` against that pipeline, and the slimmed version deletes all of those nodes. Restore filters node lists by name and silently ignores misses, so training would quietly run with no loss bridge and no metrics. The inference variant now lives in `dinomaly_cir_lentils.yaml` (which also matches its `metadata.name`), and `dinomaly_cir.yaml` is back to the training preset. There is a new guard test that fails whenever a trainrun's node lists or checkpoint monitor name nodes the referenced pipeline does not have; it immediately caught one pre-existing case, a stale `sample_metrics` entry in `rx_statistical.yaml`, now removed.

**Your pitfall 1 resolved itself this morning:** core 0.12.1 released at 07:10 with cubert-hyperspectral/cuvis-ai-core#59 in it. The uv source and the git lock entry are gone; floors are `cuvis-ai-core>=0.12.1` and `cuvis-ai-schemas[full]>=0.9.0`, and the lock is registry-only again (torch/torchvision still on the cu128 index).

**The 0.12.1 floor drags in a real migration.** Core 0.12.1 requires schemas 0.9.0, which removed `DataSplitConfig.leakage_check` (extra=forbid) in favor of the typed `constraints` list, and the old always-on default guard is gone: absent constraints now means no checks at all. Every config with an inline splits block therefore declares `no_split_overlap` explicitly now: warn where `leakage_check: warn` stood (the 8 adaclip/drcnn trainruns), error where the old default guard silently applied (adaclip_baseline, deep_svdd, data/lentils, data/tracking_cap_and_car). The three splits_path-only CuvisNEXT presets are deliberately untouched: `_effective_splits` treats constraints as file-owned and ignores inline ones when splits_path is set, so declaring them there would be dead config. Tests assert the declared constraint per config and reject any leftover `leakage_check` key anywhere under configs/.

**Security scan:** the rebase onto main brought the gitpython 3.1.59 floor and the patched aiohttp/cryptography locks, so that job goes green as a side effect. Local pip-audit also flagged pymdown-extensions 10.21.2 (three PYSECs; docs-only dep, outside the CI scan scope since CI syncs only `--extra dev`); bumped to 11.0.1 and the strict docs build still passes.

Also applied your own pitfall 6: the manifest regression test anchors on `__file__` now.

Verified locally on the branch: ruff, full `pytest -m "not slow and not gpu"` (1277 passed), the dep_compat floors audit script (0 findings), pip-audit with CI's ignore flags (clean), `mkdocs build --strict`. One thing worth confirming from your side: `plugin_pin_bump.yml` only runs when the App credentials are present, and the builtin tag now relies on it to track releases, so worth checking the App id secret is actually set on this repo.

nghorbani (approved):
Manifest fix verified end to end; floors, constraints migration and the dinomaly preset split are on the branch and CI is green. Details in my earlier comment.

PR #86, opened 2026-08-31
Add unet plugin manifest to default catalog

rajanish4:
## What
Adds `cuvis_ai/configs/plugins/unet.yaml` — a plugin manifest for **cuvis-ai-unet** (DynUNet segmentation) — to the bundled default catalog.

## Why
The catalog already ships manifests for `augment`, `dinomaly`, `adaclip`, `detr`, `sam3`, `bytetrack`, etc., but **not `unet`**. Any pipeline that declares `plugins: [unet]` (e.g. a DynUNet shell/kernel segmentation pipeline) fails to load in cuvis.next / `restore-*`:

```
required plugin 'unet' not found in catalog
```

even though `cuvis-ai-unet` is a released plugin (v0.2.0). Without the manifest the plugin can't resolve or auto-provision the way the other bundled plugins do.

## Change
`unet.yaml` in the git-tag form, matching `augment.yaml`:
- `repo: https://github.com/cubert-hyperspectral/cuvis-ai-unet.git`, `tag: v0.2.0`, `package_name: cuvis-ai-unet`
- capabilities copied verbatim from the cuvis-ai-unet repo's own `plugins.yaml` at v0.2.0: `DynUNet`, `DiceLoss`, `CrossEntropyLoss`, `OHEMCrossEntropyLoss`, `SegMetrics`, `SegmentationAnomalyScore`.

## Verification
With this manifest in the catalog, a DynUNet segmentation pipeline (`plugins: [cuvis_ai_builtin, unet, augment]`) resolves and builds:
```
BUILT walnut_unet_A_train.yaml -> [DataSource, Norm, Augment, DynUNet, DiceLoss, CrossEntropyLoss, SegMetrics]
BUILT walnut_unet_A_inference.yaml -> [DataSource, Norm, DynUNet]
```


nghorbani:
Good catch on the gap, and the manifest content itself was emit-accurate (verified in sync with scripts/emit_metadata --check). One blocker though: unet v0.2.0/v0.2.1 still carried the unscoped torch cu128 index pin that the 0.13.5 sweep removed from every other cataloged plugin. uv reads a git dependency's [tool.uv.sources], so cataloging that tag would have re-introduced the conflicting-torch-indexes child-env failure on non-cu128 hosts (Jetson Thor). I cut cuvis-ai-unet v0.2.2 (cuda dependency-group scoping, same pattern as augment v0.4.2; CI --locked runs drop --no-sources; aarch64 audit clean) and pushed 81096bc here: pin to v0.2.2, catalog-style header instead of the copied dev-manifest one, dropped the stale commented git-tag block, builtin self-pin to 0.13.7.

One thing to converge on later, not in this PR: builtin now ships DiceLoss/CrossEntropyLoss (same names, same tags) since cuvis-ai 0.13.x, so the picker will show both those and cuvis_ai_unet's. Worth deciding at some point whether a future unet release drops its local losses in favour of the builtin ones. Kept all six capabilities for now since your walnut pipelines are verified against them.

Merging and cutting 0.13.7 once CI is green.

PR #89, opened 2026-09-03
Add channel-selector (Gumbel + AdaCLIP) lentils tutorial notebook

rajanish4:
## What

A pedagogical `use_cases` notebook that **learns** the 3-band custom channel selector for the
lentils foreign-object dataset: a `ConcreteChannelMixer` (Gumbel-Softmax, 61 → 3) trained
through a frozen AdaCLIP detector with an IoU + distinctness loss. Follows the
cuvis-ai-notebook pattern — explicit node construction + `pipeline.connect`, inline
`PublicDatasets.download_dataset` + `convert_split_manifest` provisioning, metrics read off
the metric nodes.

It states the problem, visualizes the inputs (RGB / CIR / GT), the selected bands + weight
heatmap, the learned composite vs plain RGB, and the trained pipeline's anomaly maps /
predicted masks on test frames, then reports test-set metrics.

## Results (baked, 20-epoch full run on GPU)

The baked outputs come from an independent full re-run of this branch (Windows, RTX 5070 Ti laptop with 12 GB, 10 h 16 min, ~7 GB peak allocated). It reproduced the original bake: identical bands, metrics within 0.005 (original: pixel AUROC 0.9180, image AUROC 0.9037, IoU 0.2739, AP 0.4873).

- Selector converged to **[40, 59, 58] = 750 / 902 / 894 nm** — 2 of 3 bands match the
  published custom selector (542/902/886); the first band is seed/epoch/data-dependent, which
  the notebook documents.
- Section 8 test-set metrics over the **180 test frames**: **pixel AUROC 0.9183 · image AUROC
  0.9085 · IoU 0.2739 · avg precision 0.4875** — read pooled off the metric nodes via
  `.compute()` after one TEST-stage pass (so no per-batch running-mean artifact).

## Notes

- Outputs are baked in (7 figures + metrics), ~4.7 MB.
- Two run-time fixes from the re-run: the vendored `adaclip_tools` package switches matplotlib to Agg when the model loads (every `plt.show()` after training rendered nothing), so the training cell re-selects `%matplotlib inline` afterwards; and the AdaCLIP plugin logs every forward pass at INFO (19k lines per run), so the loguru sink keeps that module at WARNING. Both deserve plugin-side fixes in cuvis-ai-adaclip. Knobs default to the full recipe with a
  documented dry-run toggle (`MAX_EPOCHS = 1`, `CONVERT_LIMIT = 24`).
- Reuses the shared `use_cases/utils.py` renderers; no new dependencies.
- Reproduction notes captured in the notebook: cu3s→npz conversion is the slow part (~26 s/frame,
  resumable); reuse pre-converted npz via hardlinks (the npz_multi universe parser rejects `..`).

## Standard plugin-provisioning cell (all use_cases notebooks)

Setting up the plugins for this notebook by hand surfaced that the folder had no single way to do it, so every notebook in `notebooks/use_cases` now carries the same cell right after its Colab bootstrap. It holds a `PROVISION` list of (packaged pipeline, data module) pairs and resolves them through the core `provision` CLI against the manifests packaged with cuvis-ai (`cuvis_ai/configs/plugins/<name>.yaml`): `uv run provision ... --apply` locally, the same specs via `%pip` on Colab. Plugin pins are no longer typed by hand anywhere (bootstraps install only `cuvis-ai`), already-importable plugins are skipped, and the two SAM3 notebooks plus the metal-scrap one look up their manifests through `files("cuvis_ai") / "configs"` instead of `../../` paths. `notebooks/use_cases/README.md` and `docs/tutorials/running-notebooks.md` are rewritten around it. All edits are source-only; every baked output is byte-identical. The matching rule for the notebook skill is `cubert-hyperspectral/cuvis-ai-agentic-skills#13`.


nghorbani (commented):
Rebased onto main and pushed three follow-up commits on top of your three (force-push with lease; your commits are unchanged, `git range-diff` is clean):

- `style(notebooks)`: ruff format plus the import-sort and semicolon fixes so Linting & Style passes. Cell outputs, metadata and figures are byte-identical.
- `chore(deps)`: pip-audit started flagging tornado 6.5.7 and transformers 5.7.0 in `uv.lock` (new advisories since Aug 31, not caused by this PR, but Security Scanning is a required check). Re-locked to tornado 6.5.8 / transformers 5.16.1 and raised the transformers floor to the locked 5.16.1 (the dependency-floor audit requires floor == lock; a first attempt with 5.10 was flagged STALE and fixed in a follow-up commit).
- `docs(notebooks)`: content and docs pass. Source-only, except three cleared log outputs (bootstrap TqdmWarning, conversion log, TensorBoard run-dir warning) and the static text of cell 17's printed hint; all figures and the metric numbers are untouched.
  - Intro, section 6 and section 9 now say the baked run picked 750/902/894 nm and the whitepaper run 542/902/886 nm, and no longer claim that the defaults reproduce the published bands or that neighbouring picks are interchangeable (750 vs 542 nm is not a neighbour). Section 6 also notes that channels 2 and 3 sit on adjacent bands (902/894 nm) and why `DistinctnessLoss` cannot prevent that: it is a cosine repulsion on the selector rows, and two one-hot picks on adjacent bands are already orthogonal to it.
  - Prerequisites block with the footprint: ~57 GB snapshot (downloaded in full even for the dry run), ~140 MB per NPZ frame (~155 GB full / ~10 GB dry run), ~26 s per frame conversion, multi-hour training.
  - Bootstrap on the siblings' `IN_COLAB` pattern with the manifest pins (dinomaly v0.6.3, dataloader from PyPI >=0.6.2, adaclip v0.3.1), `utils.py` fetched from a pinned commit on Colab, Colab forced to the dry-run knobs. The old try/except import would pip-install into a local uv venv whenever one package was missing.
  - Paths: repo root derived from the `utils.py` location, so the dataset lands in the gitignored repo-level `data/` (an existing mirror there is reused) instead of notebook-relative `data/` and `data_npz/` (the latter is not gitignored, ~155 GB untracked).
  - Hygiene: `np.load` with context managers and without `allow_pickle`, loop-based `sample_row`, `input_channels` from the wavelength axis, `pipeline.torch_layers.eval()` before the two manual inference loops, processed-frame count in section 8.
  - Section 4 says the soft IoU on min-max-normalized scores is not the thresholded IoU reported later; section 8 says AP/AUROC are histogram approximations (`thresholds=200`) and that the IoU depends on the hand-set gate (0.24) and quantile (0.995).
  - README and docs: the notebook is listed in `notebooks/use_cases/README.md`, the running-notebooks page and the AdaCLIP and channel-selection tutorial pages; the install lines for dinomaly, sam3 and adaclip now use the git tags (none of the three is on PyPI).

Open points for you:

1. Objective vs reported metric. The selector is trained with `IoULoss(normalize_method="minmax")`, a soft IoU on per-batch min-max scores, while sections 7/8 report a thresholded IoU behind the decider gate, and normal frames give almost no gradient after min-max normalization. Keep as is (it is documented now) or change the objective? Your call.
2. Threshold provenance. Were `image_threshold=0.24` and `quantile=0.995` chosen on validation frames or on test frames? If validation, say so in section 8 and the IoU can be called a test metric; if test, the current "illustrative" wording should stay.
3. Optional, needs re-running cells 10 and 20 only: render the composites with `FixedWavelengthSelector` instead of the hand-rolled `spectral_composite` (compute in catalog nodes; `load_frame` is fine as I/O). A full re-bake is only worth it if you want figures and metrics regenerated with `eval()` in place (ViT-L has no dropout or batchnorm, so I expect no change).
4. Follow-ups, not for this PR: a wavelength-aware spacing regularizer for the selector, and moving `AnomalyAUROCMetrics` from cuvis-ai-dinomaly into cuvis-ai `metrics.py` (its own module docstring asks for it) so this notebook stops pulling the whole dinomaly stack for one metric node.

CI is re-running on the rebased branch; Linting & Style and Security Scanning should be green now.




**Notebook-convention pass** (cuvis-ai-notebook skill checklist), pushed as a fourth follow-up commit:

- Cells 22 and 25 now run through `Predictor.predict(...)` (`stage=INFERENCE` with `collect_ports` for the plotted ports; `stage=TEST` for the metric pass) instead of hand-rolled `pipeline.forward` loops. Predictor does the same eval-mode switch, device move, no_grad and per-batch Context, so the baked outputs should reproduce unchanged; its tqdm bar will show up as extra stderr on your next bake.
- Cell 11 no longer hand-draws the graph in ASCII; the bare `pipeline` cell two cells later renders it.
- The rest of the checklist already conforms: explicit constructors and `connect`, inline `PublicDatasets` + `convert_split_manifest`, metrics read off nodes with `.compute()`, INFO logging, Colab bootstrap, cuda/mps/cpu fallback, baked outputs with LF endings and no mojibake.
- For your re-bake: the per-frame IoU in the section 7 titles comes from `utils.iou`; with a TEST-stage pass it can be read off the `metrics` port of `metrics_anomaly` instead. `utils.py` also still carries the sibling notebook helpers that load pipelines and fetch data (`resolve_default_config`, `run_method_on_frame_subset`), which the convention keeps out of `utils`; pre-existing, out of scope here.

nghorbani (approved):
Rebased onto main and finished on the branch: lint and security gates, Predictor-based inference cells, standard plugin-provisioning cell across the use_cases notebooks, docs and changelog. The bake was reproduced end to end on a second machine (same bands, metrics within 0.005). The open design questions (soft IoU objective vs thresholded IoU, threshold provenance) are tracked in the review comment and can follow in a later PR.

PR #90, opened 2026-09-04
release: cuvis-ai 0.14.0 (class-attribute execution stages, optional two-stage gate, trained-pipeline inference)

nghorbani:
- **`TwoStageBinaryDecider.image_threshold` is optional and defaults to `None` (gate off).** With no gate every frame reaches stage 2; with `pixel_threshold` also unset the node produces the same decisions as `QuantileBinaryDecider` for `[B, H, W, 1]` input (tested bit for bit). A value that is not a finite number (a string, a bool) is refused by name instead of being coerced; the same check now covers `pixel_threshold`. Behaviour change only for `TwoStageBinaryDecider()` with no arguments (the old default gated at 0.5 in raw score space); every shipped preset that uses the class sets `image_threshold` explicitly. Per-frame debug logging is lazy, so a live camera path no longer pays whole-tensor `.item()` syncs when debug is off.
- **Training presets ship the two-stage decider with both thresholds unset.** `dinomaly_cir.yaml`, `dinomaly_rgb.yaml` and `adaclip_supervised_cir.yaml` (the pipelines behind the `*_cuvisnext` trainruns) replace `QuantileBinaryDecider` with `TwoStageBinaryDecider` carrying `image_threshold: null`, `top_k_fraction: 0.001`, `quantile: 0.995`, `pixel_threshold: null`; the adaclip node is renamed `decider`. A CuvisNEXT-saved run therefore exposes the two keys an operator sets after training. Runs trained before this release keep their always-on quantile decider and cannot be gated from the picker: retrain them (a 1-epoch run takes minutes). `tests/configs/test_cuvisnext_training_presets.py` guards the keys.
- **Execution stages are declared on the node class** (cuvis-ai-core 0.15.0, `EXECUTION_STAGES`). Losses, metrics, the TensorBoard sink and the artifact visualizers declare their stages on the class body; the copied `consume_base_kwargs` blocks, the `LossNode` assert and the `execution_stages` forwarding in `TopKIndices` are gone. A node's stages are exactly its class declaration: a pipeline yaml cannot move a node per instance (the inert `execution_stages: null` that older presets carried is accepted and dropped; any other value fails at load with a `TypeError` naming the node). The 15 `execution_stages: null` keys are stripped from the shipped presets.
- **Stage and lifecycle-tag audit of the builtin nodes.** `TensorBoardMonitorNode` runs in train/val/test (was ALWAYS) and creates its log directory and `SummaryWriter` on first use instead of in `__init__`, so loading a pipeline for inference writes no TensorBoard directory; it gains the `TRAINING` tag and a `cleanup()` hook. `ScoreHeatmapVisualizer`, `AnomalyMask`, `RGBAnomalyMask` and `CubeRGBVisualizer` run in val/test (they claimed INFERENCE although the TensorBoard sink is their only consumer) and gain `EVALUATION`; `PCAVisualization`, `PipelineComparisonVisualizer`, `ChannelWeightsViz`, `ChannelSelectorFalseRGBViz` and `ImageArtifactVizBase` gain lifecycle tags; `DistinctLabelCount` and `DeepSVDDCenterTracker` gain `INFERENCE` because they deliberately run there. New `tests/test_node_lifecycle_consistency.py` enforces: losses/regularizers never run at inference; a node pruned at inference carries `TRAINING`/`EVALUATION`; a `TRAINING`/`EVALUATION` node without `INFERENCE` is pruned at inference (augmentations exempt, they sit inline); an `INFERENCE` node runs there. The builtin manifest is regenerated so the exposed tags match (the regen also picks up doc summaries, an icon colour and a `caption` port that had drifted since the last regen).
- **CLI inference writes no TensorBoard artifacts for these visualizers any more** (`Predictor`, `restore-pipeline`). There is no yaml opt-in; `Predictor.predict(stage=ExecutionStage.TEST)` runs the val/test nodes over a labeled split when their output is wanted.
- **A restored channel selector no longer replays its warmup.** `ChannelSelectorBase._norm_frame_count` is a buffer (as in `PercentileNormalizer`), so the running-normalization warmup/freeze state is saved with the weights. Checkpoints written before this release carry no counter: strict loading accepts them and treats a fitted checkpoint as warm and frozen, an unfitted one keeps warming up. Measured on one Dinomaly run, re-warming had cut the clean/contaminated score separation from 0.082 to 0.009.
- `calibrate-thresholds` handles a two-stage decider whose gate is off (`image_threshold: null`): the current-preset row scores every frame and prints "gate off" instead of failing on `float(None)`.
- Relocked `tornado` 6.5.7 -> 6.5.8 and `transformers` 5.7.0 -> 5.16.1 (floor raised to match, `tokenizers` and `safetensors` follow) for the pip-audit advisories GHSA-8423-8fgw-73vq, CVE-2026-82397 and CVE-2026-9856; torch stays on the cu128 2.11.0 wheels.
- Bumped the plugin manifest pins to the releases that declare their execution stages on the class: `adaclip` v0.3.1 -> v0.3.2, `dinomaly` v0.6.3 -> v0.6.4, `cuvis_ai_inspecscrap` v0.2.4 -> v0.3.0 (the `execution_stages` constructor parameter of `MulticlassSegmentationMetrics` and `RgbLabelToClassIndex` is gone), `unet` v0.2.2 -> v0.3.0 (the same for `SegMetrics`; the profiling examples drop the file-reading source from the config instead of muting it), `augment` v0.4.2 -> v0.4.3.
- Bumped the `cuvis_ai_builtin` manifest pin v0.13.8 -> v0.14.0 so composed child environments install this release.
- Regenerated the `cuvis_ai_inspecscrap` manifest capabilities with `emit_metadata` against v0.3.0: the hand-written compact list becomes full entries with icons, port specs and doc summaries, so `emit_metadata --check` passes for it again.
- The metal-scrap notebook follows the plugin: it installs `cuvis-ai-inspecscrap` v0.3.0, drops the `execution_stages` argument from its dense-inference metrics sink and runs that pass with `Predictor.predict(stage=ExecutionStage.TEST)`, so the val/test metrics node scores the held-out frame instead of being pruned at inference.
- Added the `channel_selector_lentils` use-case notebook: a `ConcreteChannelMixer` learns the 3-band custom selector through frozen AdaCLIP on the lentils dataset; baked from a full 20-epoch run with the test metrics read off the metric nodes.
- Every `notebooks/use_cases` notebook now provisions its plugins through one standard cell (`uv run provision` against the packaged manifests, `%pip` on Colab); the Colab bootstraps install only `cuvis-ai` and manifest lookups use the packaged `cuvis_ai/configs`.
- Rewrote `notebooks/use_cases/README.md` and `docs/tutorials/running-notebooks.md` around the provisioning cell and removed the hand-written `git+` install lines.
- Requires `cuvis-ai-core>=0.15.0`.


rajanish4:
## Review

Read the full diff (decider refactor, the class-attribute stage migration, channel-selector buffer, calibrate-thresholds null-gate, presets, manifests, docs, and every test), and verified the claims against the branch. **Approving** — the engineering is solid and I couldn't find a correctness defect in the logic. The main actionable item is sequencing against the in-flight calibration work, not this PR.

### What I verified

- **Decider equivalence holds.** With the gate off and no `pixel_threshold`, the quantile and absolute branches preserve the exact old per-pixel-max semantics; `test_two_stage_decider_gate_off_matches_quantile_decider_exactly` pins it bit-for-bit against `QuantileBinaryDecider`. The lazy `logger.opt(lazy=True)` change removes the per-call whole-tensor `.item()` syncs from the live path — real win.
- **Legacy checkpoint path is safe.** `_load_from_state_dict` accepts counter-less checkpoints under strict loading and treats fitted bounds as warm+frozen; the three `TestRunningNormPersistence` cases (warm, nan-bounds-keep-warming, partial-resume-until-freeze) cover the branches.
- **Lazy TensorBoard writer contract is asserted** (`log_dir is None` before first use, exactly one run dir after). `cleanup()` is correctly wired — core 0.15.0's `pipeline.cleanup()` iterates nodes and calls it, so writers now flush deterministically on session close instead of only at GC.
- **Stage/tag audit is enforced**, not just declared — `test_node_lifecycle_consistency.py` R0–R3 check the whole builtin registry.
- Core `0.15.0` is published, so the `cuvis-ai-core>=0.15.0` floor is satisfiable. CI is green across all 10 checks.

### Coordination (the one thing to act on)

This overlaps the decider-threshold-calibration work in **#91** (and its core half **cuvis-ai-core#72**):

- `cuvis_ai/node/deciders/two_stage_decider.py` **auto-merges cleanly** — the `_optional_finite`/`forward` refactor here and the `calibrate()` method there are orthogonal, and they compose well: this PR ships the two-stage decider with `image_threshold: null` / `pixel_threshold: null`, and #91's calibration phase is exactly what fills those nulls in post-training.
- `scripts/calibrate_thresholds.py` and `CHANGELOG.md` **conflict** (both edit the null-gate report helpers / the changelog heading).

Suggested order: land this first, then rebase #91 onto it (small, two-file resolution). Nothing here blocks on #91/#72.

### Minor (non-blocking)

- `TensorBoardMonitorNode.cleanup()` doesn't call `super().cleanup()`, unlike the sibling `ToImage` override in `video.py`. Base `Node.cleanup()` is a no-op today so it's harmless, just a forward-compat smell.
- The regenerated `cuvis_ai_inspecscrap.yaml` carries a couple of "fitted in Phase 1" doc summaries — trips the no-"Phase N" convention. Source is the plugin's docstrings, so it's a cleanup in `cuvis-ai-inspecscrap` + regenerate, not here.

### Behavior changes to keep visible for operators

- Bare `TwoStageBinaryDecider()` no longer gates (was 0.5) — documented, and every shipped preset sets it explicitly.
- Pre-0.14 saved runs keep the always-on quantile decider and can't be gated from the picker without a retrain, as the changelog notes.

rajanish4 (approved):

PR #91, opened 2026-09-04
release: cuvis-ai 0.15.1 (decider calibrate(), shared threshold sweep)

rajanish4:
- `BinaryDecider`, `QuantileBinaryDecider` and `TwoStageBinaryDecider` gain `calibrate(scores, targets)`: re-fit the decider's own decision rule to F1-max on a labelled validation split, written to the live attribute and to `hparams`, so the saved pipeline yaml carries the value (the `.pt` is unchanged). This is the node half of the in-training calibration phase; the trainrun hook lands in cuvis-ai-core.
- The F1-max sweep math moved from `scripts/calibrate_thresholds.py` into the shared `cuvis_ai/node/deciders/_calibration.py`; the CLI and the new methods run the same code and report the same values.
- Calibrated thresholds are exact in float32 (the space the deciders compare in) and sit at the midpoint of the F1-max plateau instead of on a validation sample; the two-stage gate reuses the exact stage-1 statistic of `forward`.
- `BinaryDecider` calibration sweeps float32 sigmoid probabilities directly, so saturated logits no longer map to an unreachable threshold; the CLI binary mode follows.
- `CalibrationError` (a `ValueError`) names the bad input: mismatched shapes, non-finite scores, single-class targets, or a `QuantileBinaryDecider.reduce_dims` the sweep cannot honour.
- The sweeps count true and false positives for all candidates in one pass per frame (sorted scores plus `searchsorted`), so calibration on a large validation split no longer scans every pixel once per candidate.

Pairs with the trainrun hook in cubert-hyperspectral/cuvis-ai-core#72 (independent to merge; the feature is active in CuvisNEXT once both are released and the managed env is relocked). Review follow-ups landed on the branch as 0d25087.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 # Changelog
 
-## 0.15.0 - unreleased
+## 0.15.1 - 2026-09-04
+
+- `BinaryDecider`, `QuantileBinaryDecider` and `TwoStageBinaryDecider` gain `calibrate(scores, targets)`: re-fit the decider's own decision rule to F1-max on a labelled validation split, written to the live attribute and to `hparams`, so the saved pipeline yaml carries the value (the `.pt` is unchanged). This is the node half of the in-training calibration phase; the trainrun hook lands in cuvis-ai-core.
+- The F1-max sweep math moved from `scripts/calibrate_thresholds.py` into the shared `cuvis_ai/node/deciders/_calibration.py`; the CLI and the new methods run the same code and report the same values.
+- Calibrated thresholds are exact in float32 (the space the deciders compare in) and sit at the midpoint of the F1-max plateau instead of on a validation sample; the two-stage gate reuses the exact stage-1 statistic of `forward`.
+- `BinaryDecider` calibration sweeps float32 sigmoid probabilities directly, so saturated logits no longer map to an unreachable threshold; the CLI binary mode follows.
+- `CalibrationError` (a `ValueError`) names the bad input: mismatched shapes, non-finite scores, single-class targets, or a `QuantileBinaryDecider.reduce_dims` the sweep cannot honour.
+- The sweeps count true and false positives for all candidates in one pass per frame (sorted scores plus `searchsorted`), so calibration on a large validation split no longer scans every pixel once per candidate.
+
+## 0.15.0 - 2026-09-04
 
 - **Plugin model weights come from the public `cubert-gmbh` Hugging Face mirrors; no Hugging Face account or token is needed any more.** The manifest pins move to the plugin releases that resolve their weights through the cuvis-ai-core 0.16.0 registry: `sam3` v0.4.0 (SAM3 checkpoint from `cubert-gmbh/sam3`), `rtsam2` v0.4.0 (EfficientTAM from `cubert-gmbh/efficient-track-anything`, now including the 512x512 variants), `adaclip` v0.4.0 (CLIP backbone and AdaCLIP heads from `cubert-gmbh/clip` and `cubert-gmbh/adaclip`, `gdown` dropped) and `dinomaly` v0.7.0 (DINOv2 backbone from `cubert-gmbh/dinov2`). Core floor raised to `cuvis-ai-core>=0.16.0`. The cache folder names change (`models--cubert-gmbh--*`), so an existing install downloads its weights once more; `uv run download-model download <name>` provisions them ahead of an offline gRPC run.
 - **New docs page [Model Weights](docs/workflows/model-weights.md)**: which weights each plugin downloads, `download-model list` and `download` usage, provisioning for the offline child runtime, custom or private weights, and the licence of every mirrored file. `.env.example` now says `HF_TOKEN` is only needed for private or custom repos.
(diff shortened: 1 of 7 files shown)


rajanish4:
Paired cuvis-ai-core PR (the trainrun hook that calls these `calibrate()` methods): https://github.com/cubert-hyperspectral/cuvis-ai-core/pull/72

rajanish4:
Rebased onto `main` now that #90 (cuvis-ai 0.14.0, merged as `01e7b2e`) has landed. Clean, single commit on top of main.

Two files conflicted with #90 and were resolved:

- **`scripts/calibrate_thresholds.py`** — #90 added the `image_threshold: null` "gate off" handling to `_current_preset_two_stage` / `_print_report`; this PR had extracted the sweep math into `cuvis_ai/node/deciders/_calibration.py`. Resolution keeps #90's null-gate branch *and* routes through the extracted `_calib.*` helpers (`_calib.image_metrics_at`, `_calib.prf`).
- **`CHANGELOG.md`** — this PR's bullet moved into a fresh `## Unreleased` section above #90's `## 0.14.0`.

`two_stage_decider.py` merged without conflict: #90's `_optional_finite` gate refactor and this PR's `calibrate()` are orthogonal, and they compose — #90 ships the CuvisNEXT presets with `image_threshold: null` / `pixel_threshold: null`, and this PR's calibration phase is what fills those in post-training.

Verified locally on the rebased base: `ruff check` clean, and the affected suites pass (`test_two_stage_decider`, `test_decider_calibration`, `test_calibrate_thresholds`, `test_cuvisnext_training_presets` — 46 tests). CI re-running on the new head.

Pairs with the core half in cubert-hyperspectral/cuvis-ai-core#72 — both should land together.

nghorbani (approved):
Thanks Anish, the design is right: one implementation for the CLI and the in-training phase, `hparams` as the persistence path, pure-tensor tests. I reviewed it against the core `Serializable` / `serialize()` path and the three CuvisNEXT presets, and rather than bounce the small items back I pushed them as `0d25087` on your branch (rebased onto 0.15.0, so the CHANGELOG conflict is gone). Approving and releasing as 0.15.1 once CI is green.

What I changed, and why:

1. **Docstrings claimed `.pt` persistence.** All three `calibrate` docstrings said the value goes "to the fitted buffer (`.pt`)"; the code writes the attribute and `hparams`, and only the saved yaml carries it (which is what you describe in the PR text). An operator loading the preset yaml plus the trained `.pt` would get the stale thresholds and the docstring told them otherwise. Rewritten; the two-stage class docstring no longer says the operator sets the gate at deployment.
2. **Thresholds now sit at the F1-plateau midpoint and are exact in float32.** The sweep picked the highest F1-tied candidate, so the image gate was exactly the lowest anomalous frame's own score, and the pixel candidates were float64 quantiles that `forward` rounds to float32 (`torch.tensor(self.pixel_threshold, dtype=scores.dtype)`), so what the sweep scored was not always what `forward` decided. The sweeps now return `margin_*` keys next to the on-point optimum: midpoint to the largest sample strictly below, cast to float32 and re-verified for pixel/sigmoid thresholds, float64 for the image gate (which compares Python floats). Same validation F1, real margin against near-duplicate scores at deployment. The deciders and the CLI both read those keys.
3. **`BinaryDecider` sweeps float32 sigmoid probabilities directly.** `torch.sigmoid` in float32 is exactly `1.0` above logits of about 17, so a raw-space optimum mapped through sigmoid afterwards could be unreachable (RX scores get there). The sweep now runs on the probabilities `forward` compares, via the same torch op (`sigmoid_float32`), for the CLI too.
4. **Stage 1 is shared.** `frame_image_score` / `topk_count` in `_calibration` hold the exact statistic `forward` computes; `forward`, `calibrate` and the CLI call it, so a calibrated gate is bit-exact at runtime instead of numpy-mean-vs-torch-mean.
5. **`CalibrationError`.** One exception type for shape mismatch, non-finite scores (a NaN became a NaN threshold that `_optional_finite` refuses on reload, i.e. an unloadable yaml), single-class targets, and a `QuantileBinaryDecider.reduce_dims` that keeps a real channel axis out of the quantile (resolved against the tensor shape, so `[1, 2]` with C=1 still calibrates). `None` stays reserved for deciders that do not implement calibration. cubert-hyperspectral/cuvis-ai-core#72 needs to catch it around `decider.calibrate(...)` so a guard never aborts a finished training before the save; I left a comment there.
6. **One-pass sweep.** The 256-pass loop over the full `[N, H*W]` array is replaced by one sort per frame and class plus `searchsorted` against the candidate edges. Identical tp/fp counts (parity test against the old loop on random data with ties), and calibration on a large val split no longer scans every pixel once per candidate. F1 ties resolve to the highest candidate in every sweep, then the margin steps down.
7. **Report** carries the joint optimum and the on-point values, so the trainrun log shows when the joint pixel F1 beats the conditional one. The conditional choice stays, as you had it.
8. **Tests**: 26 new (sweep helpers, parity, float32 exactness, saturation, the both-`None` preset shape, pipeline yaml round trip, CLI-vs-method parity).

Design notes for a follow-up, not for this release:

- A normal-only validation split raises `CalibrationError`, core skips, and the pipeline ships gate-off plus the 0.5 % quantile fallback. A normal-only calibration path (gate at the max normal image score plus margin, pixel threshold at a high normal quantile) is standard AD practice and would cover the unsupervised presets; your call on how you want to shape it.
- After calibration the val metrics reported from the same split are optimistic; one sentence in the training docs when the core hook lands.
- `reduce_dims` on `TwoStageBinaryDecider` is stored and serialised but unused in `forward` (pre-existing).

PR #100, opened 2026-09-11
release: cuvis-ai 0.16.4 (runnable quickstart, cookbook links removed, noindex for old docs versions)

nghorbani:
## Problem

The public docs (and `main`) link to `cuvis-ai-cookbook` in 36 places across 13 pages, including the blood-perfusion tutorial's "Run the example" link. That repo is private, so every one of those links is a 404 for anonymous readers -- including the notebook link that should point at `notebooks/use_cases/blood_perfusion.ipynb`, which is public and already correctly linked elsewhere.

Separately, all 36 published documentation versions on `docs.cuvis.ai` are fully search-indexable: no `robots.txt`, no `noindex`, an empty "outdated version" banner (Material renders it empty without an `overrides/main.html` block), and canonicals pointing at the wrong host (`github.io` instead of `docs.cuvis.ai`). That's how an old version surfaced in search with a dead cookbook link instead of the current docs.

## Docs content

- Replace every `cuvis-ai-cookbook` link with the public equivalent (packaged pipeline/trainrun YAML run via `restore-pipeline` / `restore-trainrun`, or a notebook) where one exists; drop the 17 gRPC client-example links and the "clone the cookbook" info boxes entirely (no public gRPC example scripts exist).
- Fix the AdaCLIP tutorial and monitoring page: three cookbook trainrun YAMLs that looked like drop-in replacements (`pca_adaclip_baseline`, `concrete_adaclip`, `drcnn_adaclip`) carry no `pipeline:` reference and aren't runnable; use the packaged pipeline YAMLs and the one trainrun that's actually wired (`drcnn_adaclip_trainrun.yaml`) instead.
- Fix the quickstart: `download-data` no longer exists (now `dataset download`), the RX trainrun's `training:` block makes `restore-trainrun` take the gradient path even with no loss nodes, and the RX pipeline isn't "pre-trained" -- `restore-pipeline` statistically initializes it from the sample cube on the spot. Quickstart now uses the `StatisticalTrainer` recipe from the statistical-training doc.
- Fix two notebook links that pointed at the `notebooks/` directory instead of the specific file.
- Review round: the quickstart is now one verified flow: `uv run provision` for the `cu3s` data module (a fresh `uv sync` does not install `cuvis-ai-dataloader`), `dataset download lentils`, inspect the RX pipeline, `restore-trainrun --mode train` on the packaged statistical-only trainrun, then `restore-pipeline` on the fitted result; every command names `Lentils_000.cu3s` (there is no `Demo_000.cu3s`; same fix in `statistical-training.md` and the gRPC client example). Neither previous quickstart step ran on the released stack: the packaged RX yaml has no fitted statistics, and a cu3s module refuses to fit without a declared split, which the statistical-training recipe now declares. Commands for pipelines under the packaged `configs/` tree drop `--plugins-dir cuvis_ai/configs/plugins`: core discovers that catalog itself and rejected the duplicate manifests, so all 13 documented commands failed (core fix to follow separately). `configs/trainrun/rx_statistical.yaml` loses its `training:` block (gradient path, empty optimizer), 16 more yamls get forward-slash output paths, and `TensorBoardMonitorNode` closes its writer through `atexit`, so `restore-trainrun` exits after "Complete" instead of hanging in `__del__`.
- Fix `pyproject.toml`'s Documentation URL (wrong host/repo name) and `CONTRIBUTING.md`'s stale IA path.
- Fix `output_dir: outputs\<name>` (a literal backslash directory on Linux) to `outputs/<name>` in the 12 trainrun YAMLs that now have public doc links pointing at them.
- New `tests/docs/test_no_private_repo_links.py`: no `cuvis-ai-cookbook` reference in any published doc, and every `github.com/.../cuvis-ai` blob or tree link resolves to a real path in the repo.

## SEO

- `mkdocs.yml`: `site_url` now `docs.cuvis.ai` (was the `github.io` host); mike's `canonical_version=latest` plugin option makes every future build's canonicals, sitemap and `llms.txt` point at `/latest/`.
- `overrides/main.html`: fill in the outdated-version banner block that Material renders empty without one.
- `scripts/docs_postprocess_gh_pages.py`: post-deploy transform of the `gh-pages` checkout. Validates the tree (single `latest` alias, on the highest version, dirs present) before writing anything; injects noindex + the banner into every non-latest version's HTML; removes `llms.txt`/`llms-full.txt`/`sitemap.xml(.gz)` from old versions (they can't carry noindex and only advertise stale URLs); writes a root `robots.txt` pointing at `/latest/sitemap.xml`. Byte-based throughout so it never rewrites line endings. `tests/scripts/test_docs_postprocess_gh_pages.py` covers idempotency, version transitions, validation failures, and `--check` mode; also dry-run and real-run verified against the actual 37-version `gh-pages` branch.
- `.github/actions/deploy-docs`: composite action shared by the release workflow and the manual redeploy workflow (previously duplicated), running `mike deploy`/`set-default` without `--push`, then the post-process step against a worktree, then a single push, plus a build assertion that the deployed version canonicalizes to `docs.cuvis.ai/latest/`.

- Review round: `_remove_noindex` also strips the newline and indentation the injector wrote, so an alias flip is a byte-exact round trip (tested, including a yanked release); the guard test ignores `#fragment` / `?query` in blob links; `CHANGELOG.md` gains the `## Unreleased` entry.

## Follow-up (separate, after merge)

- Prune the 20 versions below 0.13 with `mike delete`.
- Trigger the manual redeploy workflow for 0.16.2 to backfill noindex/banner into the remaining old versions.
- Submit Search Console removal requests for the retired version prefixes and the `/latest/` sitemap.

## Verification

- `uv run mkdocs build --strict` -- clean.
- `uv run pytest tests/scripts/test_docs_postprocess_gh_pages.py tests/docs/test_no_private_repo_links.py tests/configs` -- 185 passed, 9 skipped (expected: statistical-only trainruns, symlink test on Windows).
- `uv run ruff format --check .` / `uv run ruff check .` -- clean.
- `actionlint` -- clean.
- `docs_postprocess_gh_pages.py --check` and a real run, both dry-run and applied, against the actual `origin/gh-pages` branch in a detached worktree (not pushed).


rajanish4:
## Review

Reviewed the full diff and ran the verification battery locally (isolated worktree at the PR head, based on latest main). The script/tests/CI-action side is solid — I could not break it, including against the real 36-version `gh-pages` tree. One content fix before merge, plus a changelog entry.

### 1. `Demo_000.cu3s` doesn't exist — the quickstart demo fails on first run

The HF repo `cubert-gmbh/XMR_Lentils` ships exactly `Lentils_000.cu3s` (920 MB, matching the quickstart's "~1.0 GB" claim). There is no `Demo_000.cu3s`. But:

- `docs/get-started/quickstart.md:40` — the "Quick Demo" `restore-pipeline` command (in the section this PR rewrote) uses `cu3s_file_path=data/Lentils/Demo_000.cu3s` → FileNotFoundError right after the download step produces `Lentils_000.cu3s`.
- `quickstart.md:56` — the new `StatisticalTrainer` snippet added here → same wrong file (faithfully copied from `docs/workflows/statistical-training.md:27`, which has the same pre-existing error, untouched by this PR).
- The page is now internally inconsistent: download → `Lentils_000.cu3s`, demo/train → `Demo_000`, inference step → `Lentils_000`.

This is exactly the "docs must be runnable for anonymous readers" class of bug the PR targets, and the runnable-examples test that would catch it is `--runslow`-gated. Suggest `s/Demo_000/Lentils_000/` in quickstart (×2), and ideally `statistical-training.md:27` in the same pass.

### 2. No CHANGELOG entry

CHANGELOG.md is untouched, but this changes user-visible behavior (docs URL/canonicals, robots.txt, noindexing every old version, new script + deploy action). Worth a line per the per-PR changelog convention.

### What I verified — all green

Local (ruff clean; `pytest tests/scripts/test_docs_postprocess_gh_pages.py tests/docs tests/configs` → 365 passed, 24 skipped, all expected; `mkdocs build --strict` passes).

Against the **real** `origin/gh-pages` (detached worktree, nothing pushed):
- `--check` → exit 1 (2308 html would change, 130 removed, robots.txt would change).
- real run → second `--check` → exit 0. **Idempotent on real data**, not only fixtures.
- Spot checks: `0.10.0/index.html` has exactly one marked noindex + the banner; `0.16.2` (latest) has zero markers; `llms*.txt`/`sitemap*` stripped from old versions only; `robots.txt` byte-exact; the `latest` alias never touched.

Every runnable claim checked against the installed APIs: `dataset download {lentils,blood_perfusion}` (lowercase aliases + `download` subcommand exist), `CuvisPipeline.load_pipeline`, `StatisticalTrainer(pipeline=…, datamodule=…).fit()`, `save_to_file`, `Cu3sDataModule(cu3s_file_path=…)` — all match.

Design worth calling out as correct: validate-then-write with a dry pass (one unparseable HTML file anywhere aborts before a single byte is written — confirmed by test); fail-closed validation (two `latest` aliases / `latest` not on the highest version both rejected); and `robots.txt` deliberately not `Disallow`-ing old versions (correct — a robots-blocked page can't be crawled to see its noindex). Both deploy workflows share the `docs-deploy` concurrency group + container, so gh-pages pushes serialize.

### CI

The red **Documentation Build** is an `archive.ubuntu.com` mirror outage during `apt-get install` (many "Unable to connect" fetch errors), not a PR defect — my local `mkdocs build --strict` passes. Please re-run it before merge.

### Minor (non-blocking)

- `_remove_noindex` leaves the `\n      ` the injector added, so each outdated→latest→outdated flip of a version accretes one blank line after `<meta charset>`. Cosmetic, bounded by alias flips.
- `REPO_LINK_PATTERN` in the guard test doesn't strip `#fragments`, so a future `blob/main/file.md#L10` link would false-fail the `exists()` check. No such link today.
- Pre-existing, out of scope: `grpc-workflow.md:399` also says `data/lentils/Demo_000.cu3s` (wrong case + name) in illustrative code.

**Recommendation:** fix the `Demo_000` → `Lentils_000` occurrences + add a changelog line, re-run the flaked Documentation Build, then good to merge.

nghorbani:
Thanks for the thorough pass. Both items are in, plus what fell out when I ran the quickstart end to end instead of checking the API names.

**Pushed** (branch rebased on 0.16.3, head `09938f3`; commits on top of the original one):

- `docs: point the quickstart demo at Lentils_000.cu3s` (quickstart x2, `statistical-training.md`, and the gRPC client example, which also had `data/lentils/`)
- `docs: provision the cu3s data module before the quickstart demo`: a fresh `uv sync` does not install `cuvis-ai-dataloader` and `restore-pipeline` runs in-process, so the demo failed before reaching the data; both getting-started pages now run `uv run provision` first
- `configs: make the rx_statistical trainrun statistical-only` and `docs: train the quickstart pipeline with the packaged trainrun`: neither previous quickstart step ran on the released stack. `restore-pipeline` on the packaged RX yaml has no fitted statistics (`MinMaxNormalizer requires statistical_initialization()`), and the `StatisticalTrainer` snippet fitted a cu3s module without a declared split, which dataloader 0.6.3 refuses. The flow is now inspect -> `restore-trainrun --mode train` (statistical-only, ~30 s, writes `outputs/rx_statistical/trained_models/RX_Statistical_restored.{yaml,pt}`) -> `restore-pipeline` on the fitted yaml. The statistical-training recipe declares its train split with a `file_indices` selector.
- `docs: drop the packaged plugins dir for packaged pipelines`: core's `_discover_plugins_dirs` already finds `configs/plugins` next to a packaged pipeline and then appends the same directory again from `--plugins-dir`, so `_build_catalog` rejects every manifest as a duplicate plugin name. All 13 documented commands of that shape failed on core 0.17.2 (OS-independent; the reviewer battery checks API names, not the commands). Packaged-pipeline commands lose the flag; `outputs/`-style paths keep it. The proper fix is a resolved-path dedupe in core, tracked as ALL-6466.
- `monitor: close the TensorBoard writer through an exit hook`: `restore-trainrun` logged "Complete" and then never exited. `TensorBoardMonitorNode.__del__` closed the writer at interpreter shutdown and blocked forever in TensorBoard's event-writer `stop()`. The writer now registers an `atexit` hook; `__del__` skips the close while finalizing. Two tests.
- `configs: forward slashes in the remaining output_dir and dirpath values` (13 pipeline presets, 3 trainruns), same fix as the trainrun `output_dir`s in this PR
- `docs: remove the noindex tag together with its indentation` (byte-exact round trip, tests for both head anchors and a version regaining `latest`) and `tests: ignore fragments and queries in repo blob links`, your two minor items
- `changelog:` `## Unreleased` section covering all of the above

**Verified locally** (worktree venv, Windows, Cuvis SDK 3.5.3): every quickstart command against `data/Lentils/Lentils_000.cu3s` (14 measurements) in the documented order: provision (installs `cuvis-ai-dataloader 0.6.3` + `cuvis 3.5.3.2`), inspect, `restore-trainrun --mode train` (3 statistical nodes fitted, test split scored, exits in 50 s), `restore-pipeline` on the fitted yaml (14 frames, 13 FPS, clean exit), and the Python recipe from `statistical-training.md`. `ruff format --check` / `ruff check` clean; the pre-push gate's `pytest -m "not slow and not gpu"`: 1659 passed, 12 skipped (the docs, configs, post-processing script and monitor tests also run on their own: 323 / 87 / 162 passed); `mkdocs build --strict` clean.

**Not exercised:** the blood-perfusion first-pipeline command (11 GB dataset; only the flag removal changed there, same code path as the verified RX command); the runnable-examples `--runslow` tests.

**CI on the new head:** all nine checks pass on `09938f3`, including Documentation Build (3m56s) and the Test Suite (4m55s). The earlier Documentation Build failure was the apt mirror (`Err:17 http://archive.ubuntu.com/ubuntu noble/main amd64 libfribidi0`), not the PR.

Left for the merge: whether this is cut as 0.16.4 (one heading rename in the changelog), and the post-merge list in the description.

rajanish4:
## Re-review of the update

Both earlier findings and both minors are addressed, and you caught three more real bugs on top — nice. Re-verified on the new head (`09938f3`, rebased onto latest main): all 9 CI checks green, and I re-ran the changed surface locally — 249 passed, ruff clean, and the statistical-only-trainrun skip count went 8→9, confirming `rx_statistical` now validates as statistical-only.

Confirmed resolved:
- Quickstart uses `Lentils_000.cu3s` throughout (the file the dataset actually ships); the whole flow (`provision` → `dataset download` → `restore-trainrun --mode train` → `restore-pipeline`) is now internally consistent.
- `rx_statistical.yaml` trainrun is statistical-only — no more empty-optimizer gradient path.
- The `--plugins-dir` duplicate-catalog failure is a great catch; that one broke every packaged-pipeline command and I'd missed it (I checked signatures, not the end-to-end run).
- `TensorBoardMonitorNode` shutdown deadlock: the `atexit` + `sys.is_finalizing()` split is the right pattern, and it closes the loop on the `cleanup()`/`__del__` path from #90.
- noindent-strip and the fragment/query handling in the guard test.
- CHANGELOG covers all of it.

One minor, non-blocking: `atexit.register(self.cleanup)` is never paired with `atexit.unregister` in `cleanup()`, so it holds a strong reference to the node for the process lifetime — a long-lived process that creates many `TensorBoardMonitorNode`s (e.g. a gRPC server across training sessions) accumulates them until exit, and calling `.log()` after an explicit `cleanup()` registers a second callback. Harmless for the CLI target here; a one-line `atexit.unregister(self.cleanup)` at the end of `cleanup()` would close it if you want to.

LGTM to merge once you're happy with (or defer) that nit.

PR #106, opened 2026-09-22
release: cuvis-ai 0.17.1 (per-frame running bounds, read-ahead in the wizard preset, dataloader v0.8.0)

nghorbani:
Release candidate for cuvis-ai 0.17.1, open for review; the merge itself waits for the items under "Before merging". Four things ship in it:

1. **`FixedWavelengthSelector` and `PercentileNormalizer` count frames, not batches, in `running` mode** (ea8347d1). One shared per-frame step in `cuvis_ai/node/_running_bounds.py`; batch-1 runs and their checkpoints are unchanged.
2. **A contract test on the cu3s reader levers a shipped trainrun may set** (9dc2af1a, 807e8a3f): `num_workers` must be 0, `sdk_cuda` unset or true, `read_ahead` within 0 and 8, and the dataloader manifest at or above the release that introduced each lever.
3. **The CuvisNEXT wizard preset reads one frame ahead of the model step** (ebc25522): `read_ahead: 1` in `dinomaly_custom_cuvisnext.yaml`, the dataloader manifest at v0.8.0 (released 2026-09-22) and the built-in manifest at v0.17.1, the tag this PR's squash commit will carry. Measured on the tester dataset (RTX 5070 Ti laptop): training step 0.24 to 0.165 s, the wizard's run 631 to 488 s, identical AUROC, the same peak device memory as before because depth 1 reads on the recording's own SDK handle. Details in the changelog entry.
4. **Security floor `gitpython>=3.1.62`** (e1585f19, 520e7743): three advisories fixed in 3.1.60 (arbitrary path reads through `--no-index` in the diff API, quadratic backtracking in the author regex, git-directory impersonation through tracked files). The lock also moves anyio, jupyter-server and soupsieve past their September advisories; those are dev-only. The same commit drops `PercentileNormalizer._per_frame_percentile`, which the per-frame refactor had left without a caller (the two lines Codecov reported uncovered).

**Review focus.** The node change and its tests (`tests/node/test_running_bounds.py`, the regression cases in the selector and normalizer suites), the contract test's rules, and the preset comment, which is what a reader of the YAML gets.

**Before merging**

- The changelog stamp (`## Unreleased` to `## 0.17.1 - <date>`) is the last commit before the squash; every check is green as of 520e7743.
- Squash with the `release: cuvis-ai 0.17.1 (...)` title, tag `v0.17.1` on the squash commit; the release workflow publishes to PyPI and creates the GitHub release.
- The 8 GB acceptance run (V0, V7a, V7, V8a on an 8 GB laptop, instructions in the dev-docs spike README) confirms the gain on the slower card; `cuda_cubes` on top of the read-ahead is decided from that run and would come in a follow-up release.


cuvis_ai/configs/plugins/cuvis_ai_builtin.yaml
```diff
@@ -9,7 +9,7 @@
 name: cuvis_ai_builtin
 package_name: cuvis-ai
 repo: https://github.com/cubert-hyperspectral/cuvis-ai.git
-tag: "v0.16.0"
+tag: "v0.17.1"
```
rajanish4: This PR and #107 both release as 0.17.1. #107 stamps `## 0.17.1 - 2026-09-22` and says this one becomes 0.17.2, but the title, the description ("tag v0.17.1 on the squash commit"), the changelog line and `tag: "v0.17.1"` here all still say 0.17.1.

If #107 is tagged v0.17.1 first and this merges unchanged, this pin resolves to #107's release, which has no `_running_bounds.py`, so CuvisNEXT training children would run the old batch-counting selector while the host has the fix. Can we settle the order and renumber (title, description, changelog, pin) before either merges? Both PRs also edit the same `uv.lock` hunks, the gitpython floor and CHANGELOG, so the second one needs a rebase anyway.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+- **The built-in plugin manifest pins cuvis-ai v0.17.1**, so a CuvisNEXT training child composed from it runs this release's nodes (the per-frame running bounds below) instead of 0.16.0's.
```
rajanish4: Separate from the numbering: pinning the builtin manifest to this package's own next tag means the tag doesn't exist until after the merge. I hit this on a branch with this PR's content (ebc25522): CuvisNEXT failed with "Tag v0.17.1 not found" until I switched the manifest to `path:`.

Anyone on main between the merge and the tag push hits the same failure, and on every dev branch the child environments run the last release's nodes instead of the checkout's. Could the builtin resolve from the installed distribution instead (the package root as `path`, or the version from `importlib.metadata`)? Then no release has to remember the bump, and the pin test in #107 isn't needed.

cuvis_ai/node/_running_bounds.py
```diff
@@ -0,0 +1,147 @@
...
+        return RunningPhase(
+            count=count, in_warmup=count <= self.warmup_frames, should_update=should_update
+        )
+
+
+def per_frame_quantiles(data: Tensor, low: float, high: float) -> tuple[Tensor, Tensor]:
+    """Per-frame, per-channel quantiles of a BHWC tensor as two ``[B, C]`` float tensors."""
+    batch, channels = data.shape[0], data.shape[-1]
+    flat = data.reshape(batch, -1, channels).float()  # quantile() requires float/double
+    return torch.quantile(flat, low, dim=1), torch.quantile(flat, high, dim=1)
```
rajanish4: Two points on the per-frame quantiles:

- `torch.quantile` is called twice, so the frame is sorted twice. One call with `q=torch.tensor([low, high])` returns both. On a 1000x1080x3 frame on CPU I measured 648 ms per frame for the two calls and 327 ms for one (CUDA: about 3 ms either way).
- `running_normalize` computes them for every frame, including frames after the bounds froze and warmup ended, where the result is thrown away. A CPU-only venv doing live inference pays about 0.65 s per frame for that after the freeze. Could the step decide first and skip the quantile once `count > max(warmup, freeze)`?

Smaller: `step()` calls `.item()` and `isnan().any()` per frame (lines 82 and 85), two host syncs per frame; it doesn't matter at batch 1. All of this predates the PR, but the PR rewrites exactly this function, so now seems a good time.

tests/configs/test_trainrun_reader_flags.py
```diff
@@ -0,0 +1,149 @@
...
+LEVERS = ("read_threads", "sdk_cuda", "cuda_cubes", "read_ahead")
+# The dataloader release that introduced each lever; an older plugin dies on the keyword.
+LEVER_FLOORS = {
+    "read_threads": Version("0.7.0"),
+    "sdk_cuda": Version("0.7.0"),
+    "cuda_cubes": Version("0.7.0"),
+    "read_ahead": Version("0.8.0"),
+}
+LEVERS_FLOOR = min(LEVER_FLOORS.values())
+MAX_READ_AHEAD = 8  # every frame in flight is a whole cube; the dataloader refuses more
```
rajanish4: Nit: `MAX_READ_AHEAD`, `LEVER_FLOORS` and the `num_workers` / `sdk_cuda` rules copy the dataloader's own validation. If the dataloader changes its cap or adds a constraint, this test stays green on the stale copy. Where CI has the plugin installed, could it validate `data.params` through the dataloader's own option parser, with this table only as a fallback?

rajanish4 (commented):
Reviewed the per-frame running-bounds change, the reader-lever contract test and the preset, and ran the suite and a restore check locally.

**Tests** (Windows, Python 3.13, fresh venv from this lock, `uv sync --locked --all-extras`, `pytest tests/ -m "not slow and not gpu"`): 1713 passed, 12 skipped, 0 failed.

**Restore / frame-order check** on the running path: `FixedWavelengthSelector` and `CIETristimulusRGBSelector` into `PercentileNormalizer` (all `running`, warmup 10, freeze 20), saved with `save_to_file` and restored with `load_pipeline`:
- checkpoint trained on 30 frames: a frame gives bit-identical output whether it is the first after the restore or comes after 15 very bright / very dark frames, and also as the last frame of a batch of 4; the bounds don't move;
- selector checkpoint without the counter key: loads frozen through the legacy hook, same result;
- checkpoint saved after 12 frames, or the yaml loaded without weights: the output still depends on frame order until the freeze, as designed.

Identical results on 0.17.0 code, so the refactor keeps that behaviour. The node change and its tests look right to me. The main issue is the release numbering against #107 (inline); the other comments are smaller.

cuvis_ai/configs/plugins/cuvis_ai_builtin.yaml
nghorbani: Renumbered in 50e9462d: title, description, changelog heading `## 0.17.2 - <date>` (re-stamped at merge) and the builtin pin `v0.17.2`. Order: #107 merges first and tags v0.17.1; this branch rebases onto it (one changelog conflict; lock and pyproject taken from main, since #107 carries the same upgrades), then merges and tags v0.17.2.

CHANGELOG.md
nghorbani: Agreed on the shape of the problem; not for this release. The schemas know two plugin sources (git tag, local path) and the composer has a notion of the installed distribution only for core itself, so this is a cuvis-ai-schemas plus cuvis-ai-core feature: a source bound to the installed package (`package_name` plus `importlib.metadata.version`), emitted by the composer as `cuvis-ai==<installed>` the way it pins core. I am opening a ticket for it. Until then the pin test in #107 ties the tag to the changelog heading, and the release train creates the tag within minutes of each squash, which is the gap you hit on ebc25522.

cuvis_ai/node/_running_bounds.py
nghorbani: Done in 34ca8de2: one `torch.quantile` call with `q = [low, high]` per batch; quantiles only for the leading frames that are still inside the warmup or before the freeze (`RunningBounds.frames_needing_quantiles`), none afterwards; the counter is read once per batch and the NaN check happens once per `RunningBounds`. Buffers and checkpoints unchanged; the parity tests now cover a warmup before, beyond and at the freeze, bounds that never freeze, and assert the call counts.

tests/configs/test_trainrun_reader_flags.py
nghorbani: Partly, in 9261750a. The table stays the verdict: CI does not install the plugin, and the parser treats `read_threads > 1` at batch 1 without `read_ahead` as a warning where the table refuses, so a parser-else-table test would judge differently locally and in CI. Where the plugin is installed, a second test asserts the read-ahead cap matches and runs every shipped trainrun through `parse_cu3s_reader_options` (the SDK device recording patched out), skipped otherwise, so a rule the dataloader tightens shows up before a training child dies on it.

nghorbani:
Thanks, all four points handled: 50e9462d renumbers to 0.17.2, 34ca8de2 takes the quantile cost out (one call per batch, none after the freeze, one counter read), 9261750a adds the parser cross-check; the builtin-from-installed-distribution proposal becomes a schemas plus core ticket. Merge order and the rebase are in the description. Re-requesting your review.

CHANGELOG.md
nghorbani: Ticket for it: ALL-6550 (schemas + core: a plugin source bound to the installed distribution; the builtin manifest then carries no version).

nghorbani:
@rajanish4 merging this now on Nima's instruction (CI green, no re-review wait). Plan change from tonight: one release only, so this lands as 0.17.1 together with #107 (aarch64 cu130 torch fork, torchcodec dropped) and #102 (sam3 v0.5.1, core floor 0.17.3); the builtin pin stays v0.17.1, the one your stack already references. Your four points are folded in (renumbering, one quantile call per batch and none after the freeze, the parser cross-check when the dataloader is installed, and ALL-6550 for the builtin manifest from the installed package). Anything you still find on the merged diff is welcome and goes into the next patch; the 8 GB acceptance run for the read-ahead preset stays open as documented in the description.

PR #107, opened 2026-09-22
release: cuvis-ai 0.17.1 (torch from the cu130 index on aarch64 / Jetson Thor checkouts)

nghorbani:
## Why

A cuvis-ai checkout synced with `uv sync` on a Jetson Thor (JetPack 7, aarch64) installed torch 2.11.0+cu128 from the cu128 index and failed at the first CUDA kernel with "no kernel image is available for execution on the device": the cu128 index serves an SBSA (Grace / GH200) aarch64 wheel whose kernels stop at sm_120, the Thor is sm_110. CuvisNEXT's managed environment already uses the cu130 index on Jetson; the checkout path did not. Seen on the Thor devkit on 2026-09-22 with the walnut FO stack.

## What

- `pyproject.toml`: the `cuda` group's torch / torchvision index pins fork by platform. `sys_platform == 'linux' and platform_machine == 'aarch64'` resolves from `pytorch-cu130`, the De Morgan complement keeps `pytorch-cu128` (PEP 508 has no `not`). Both entries stay scoped to the `cuda` group, so a git or path consumer of this package (the `cuvis_ai_builtin` manifest in a composed child environment) inherits no index pin. The base torch / torchvision floors are declared once per fork: uv assigns one index per fork, and an unsplit base requirement next to two gated indexes fails with "conflicting indexes for package torch" (verified with uv 0.8.24 and 0.12.17).
- `uv.lock`: `uv lock --upgrade-package torchcodec`. torch 2.11.0+cu128 / 2.14.0+cu130, torchvision 0.26.0+cu128 / 0.29.0+cu130, torchcodec 0.11.1 (cu128 side) / 0.16.0 (aarch64 side), triton 3.6.0 / 3.8.0, plus the cu13 NVIDIA runtime packages for the aarch64 fork.
- Second commit: the 0.17.0 lock fails the Security Scanning job (pip-audit). anyio 4.13.0 (CVE-2026-63374, CVE-2026-64847), gitpython 3.1.59 (PYSEC-2026-3982/3983/3984), jupyter-server 2.20.0 (CVE-2026-86049), soupsieve 2.8.4 (CVE-2026-85999, CVE-2026-86000) and, in the docs extra, mkdocs-material 9.7.6 (PYSEC-2026-3864) move to 4.14.2, 3.1.62, 2.21.1, 2.9.2 and 9.7.7; the gitpython floor follows the lock. These are the same upgrades PR #106 carries, so its rebase reconciles them. `pip-audit` over all extras with the CI flags is clean.
- torchcodec forks with torch (fourth commit): `torchcodec>=0.11.1,<0.12` beside the cu128 torch 2.11, `torchcodec>=0.16.0` beside the cu130 torch 2.14, same markers as the sources. The first revision of this branch moved everything to 0.16.0 on the reading that 0.12+ accepts torch >= 2.11; the new import test then failed on Linux CI with `libnvrtc.so.13: cannot open shared object file`. Every 0.12+ torchcodec wheel on Linux links the CUDA 13 runtime (`libcudart.so.13`, `libnvrtc.so.13`, `libnvjpeg.so.13`; checked by scanning the PyPI wheels), 0.11.1 x86_64 links none of them, so torchcodec has to follow the CUDA major of the torch it sits beside. Children inherit the forked floors.
- `tests/test_pyproject_torch_sources.py`: the sources are a group-scoped two-entry fork whose markers parse and partition Jetson / Linux x86_64 / Windows, the base requirements mirror the markers, the declared indexes exist and are explicit. In the lock, the resolution markers select exactly one torch and one torchvision entry per platform and Python minor from the expected index, the cu130 entries carry a `manylinux_2_28_aarch64` wheel and the cu128 entries an x86_64 and a `win_amd64` wheel for each of cp311, cp312 and cp313, torchcodec has one PyPI entry per environment (0.11.x beside cu128, 0.12+ beside cu130), and `import torchcodec` plus `torchcodec.decoders.VideoDecoder` load against the installed torch. `tests/plugins/test_builtin_manifest_pin.py` ties the `cuvis_ai_builtin` tag to the newest CHANGELOG heading; the pin had stayed at v0.16.0 through six releases.
- `.github/workflows/no-local-sources.yml`: flattens list-form sources before looking for a `path`. A fork entry carrying a `path` slipped past the `isinstance(v, dict)` check; verified against a poisoned copy of this pyproject (third commit).
- `cuvis_ai/configs/plugins/cuvis_ai_builtin.yaml`: pin v0.16.0 -> v0.17.1.
- `docs/get-started/installation.md`: Python 3.11 to 3.13 (the page said 3.10 minimum). The Jetson / aarch64 section states that the cu130 wheels need a CUDA 13 driver (JetPack 7, driver 580 or newer), that JetPack 6 is not covered for that reason, and that an aarch64 host still on a CUDA 12 driver reinstalls torch / torchvision from the cu128 index by hand and runs with `uv run --no-sync`. An earlier revision of this branch documented `uv sync --no-default-groups` as the escape hatch; that was wrong: the base requirement binds aarch64 Linux to cu130 whichever groups are synced.
- `CHANGELOG.md`: `## 0.17.1 - 2026-09-22`.

## Verified

- Windows: `uv sync --locked --all-extras`, `ruff format --check`, `ruff check`, `interrogate`, `mkdocs build --strict`, the 50 cases of `tests/test_pyproject_torch_sources.py` plus the manifest pin test, the non-slow non-GPU suite through the pre-push gate.
- Lock, statically: the cu130 fork lists cp311 / cp312 / cp313 `manylinux_2_28_aarch64` wheels for torch 2.14.0+cu130 and torchvision 0.29.0+cu130, torchcodec 0.16.0 has aarch64 wheels for the same three. Open point for the device proof: torchcodec 0.16 loads its image library at import and that library needs `libnvjpeg.so.13`, which no package in the aarch64 fork provides and torch does not preload; on a JetPack 7 device the CUDA 13 toolkit is expected to supply it.
- Jetson Thor with this exact lock: not yet run. The gate before the tag is the drop-in on the device (this `pyproject.toml` + `uv.lock` into the checkout there, `uv sync --frozen`, `sm_110` in `torch.cuda.get_arch_list()`, `import torchcodec`, CuvisNEXT starting the server from that venv and composing child environments against the cu130 index only). It runs once the checkout's owner frees the directory.

Release 0.17.2 (PR #106, per-frame running bounds, read-ahead preset, dataloader v0.8.0) follows on top of this one.


rajanish4 (commented):
Reviewed the torch / torchvision / torchcodec fork, the lock, the guards and the docs, and ran the suite locally.

**Tests** (Windows, Python 3.13, fresh venv, `uv sync --locked --all-extras`, `pytest tests/ -m "not slow and not gpu"`): 1 failed, 1707 passed, 12 skipped. The failure is `test_torchcodec_loads_against_the_installed_torch` on a machine without FFmpeg shared libraries; it also fails when the file runs alone (details inline).

**Device data point:** on the Thor devkit (JetPack 7.0, L4T R38.2.2), torch 2.14.0+cu130 cp311 from the cu130 index runs our RF-DETR segmentation pipeline on the GPU (sm_110); the cu128 wheel fails with "no kernel image". torchcodec was not exercised there, so the libnvjpeg open point stays open.

The cu130 direction is right. The torchcodec pins need another look before this ships, because they break installs that don't go through this lock (inline).

pyproject.toml
```diff
@@ -25,16 +25,23 @@ dependencies = [
...
+    # torch and torchvision are declared twice with complementary markers on purpose: the
+    # `cuda` group's index pins below fork on the same markers, and uv can only assign an
+    # index to a requirement that lives inside one fork (an unsplit requirement next to two
+    # gated indexes fails with "conflicting indexes for package torch").
+    "torch>=2.11.0 ; sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "torch>=2.11.0 ; sys_platform != 'linux' or platform_machine != 'aarch64'",
+    "torchvision>=0.26.0 ; sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "torchvision>=0.26.0 ; sys_platform != 'linux' or platform_machine != 'aarch64'",
+    "torchcodec>=0.16.0 ; sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "torchcodec>=0.11.1,<0.12 ; sys_platform != 'linux' or platform_machine != 'aarch64'",
```
rajanish4: The torch/torchcodec pairing only holds inside this checkout's lock. For everyone else torch is still `>=2.11.0` with no cap, and `torchcodec<0.12` now holds them to 0.11.

PyPI and the cpu index both serve torch 2.14.0 today, so a plain `pip install cuvis-ai` on Windows or x86_64, or a composed child on a CPU-only host (the "cpu on a CPU-only host" case from the sources comment), gets torch 2.14 with torchcodec 0.11.1. That's the ABI mismatch this PR says fails at import. Before this change those installs got the newest torchcodec next to the newest torch.

The torch floors would need to fork as well (`torch>=2.11,<2.12` beside `torchcodec<0.12`, `torch>=2.14` beside `torchcodec>=0.16`); right now `test_base_requirements_are_split_along_the_fork_markers` blocks exactly that by requiring identical floors on both forks.

pyproject.toml
```diff
@@ -25,16 +25,23 @@ dependencies = [
...
-    "torchcodec>=0.11.1",
+    # torch and torchvision are declared twice with complementary markers on purpose: the
+    # `cuda` group's index pins below fork on the same markers, and uv can only assign an
+    # index to a requirement that lives inside one fork (an unsplit requirement next to two
+    # gated indexes fails with "conflicting indexes for package torch").
+    "torch>=2.11.0 ; sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "torch>=2.11.0 ; sys_platform != 'linux' or platform_machine != 'aarch64'",
+    "torchvision>=0.26.0 ; sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "torchvision>=0.26.0 ; sys_platform != 'linux' or platform_machine != 'aarch64'",
+    "torchcodec>=0.16.0 ; sys_platform == 'linux' and platform_machine == 'aarch64'",
```
rajanish4: Question: does cuvis-ai still need torchcodec? `git grep torchcodec` finds no import anywhere under `cuvis_ai/`, only the comment in `_ffmpeg_bootstrap.py`; it looks left over from before the cookbook extraction. If the plugin that decodes video declares it itself, the fork, the import test, the new upper bound dep_compat has to check against plugins, and the `libnvjpeg.so.13` open point on the Thor all go away.

docs/get-started/installation.md
```diff
@@ -45,6 +45,22 @@ cd cuvis-ai
...
+
+```bash
+uv run python -c "import torch; print(torch.__version__, torch.cuda.get_arch_list())"
+```
+
+The list must contain `sm_110`. JetPack 6 (Orin) is not covered: its CUDA 12 driver cannot load the cu130 wheels this lock resolves, and its system Python is 3.10. An SBSA host (Grace, GH200) still on a CUDA 12 driver is outside the lock as well: `uv sync` installs the cu130 build there whichever dependency groups are selected. As a workaround, replace torch and torchvision by hand after the sync and skip the re-sync when running, because every `uv sync` and plain `uv run` restores the locked cu130 build:
+
+```bash
+uv sync
+uv pip install --reinstall "torch==2.11.0+cu128" "torchvision==0.26.0+cu128" --index-url https://download.pytorch.org/whl/cu128
```
rajanish4: This reinstalls torch and torchvision from cu128 but leaves the locked aarch64 torchcodec 0.16.0 in place. By this PR's own reasoning 0.16 links the CUDA 13 runtime and belongs beside torch 2.14, so the result is the mismatch the PR fixes elsewhere. It probably needs `torchcodec==0.11.1` (from PyPI) in the same step. `--index-url` also replaces PyPI entirely; worth checking that torch 2.11's non-NVIDIA dependencies (`cuda-toolkit`, `cuda-bindings`) resolve from the cu128 index alone.

tests/test_pyproject_torch_sources.py
```diff
@@ -0,0 +1,241 @@
...
+            for entry in os.environ.get("PATH", "").split(os.pathsep)
+            if entry
+        )
+    return ctypes.util.find_library("avcodec") is not None
+
+
+def test_torchcodec_loads_against_the_installed_torch(lock: dict) -> None:
+    """torchcodec's shared library is built per torch ABI; a mismatch fails at import time."""
+    import torch
+    import torchcodec
```
rajanish4: On a Windows machine without FFmpeg shared libraries this fails instead of skipping: `import torchcodec` raises "Could not load libtorchcodec" before the skip at line 231 runs. That happens in the full `-m "not slow and not gpu"` suite and when the file runs alone, and the pre-push hook runs exactly that suite, so it blocks pushes from such a machine.

Moving the FFmpeg check and `import cuvis_ai` (which registers the DLL directories on Windows) above the torchcodec import fixes it. Also, `assert VideoDecoder is not None` can't fail; opening a tiny clip would actually exercise the library load.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## 0.17.1 - 2026-09-22
```
rajanish4: See my comment on #106: both PRs release as 0.17.1. The date here is also stamped before the Thor device check has run, so it will be stale by the time the tag exists.

pyproject.toml
nghorbani: Agreed, and rather than forking the torch floors we dropped torchcodec (6ed58ba7): cuvis-ai imports it nowhere, so the pairing problem leaves with it. `test_base_requirements_are_split_along_the_fork_markers` keeps identical torch floors on both forks, as before.

pyproject.toml
nghorbani: No. It came in with the 0.7.0 cookbook extraction so that core's `VideoIterator` could decode on the GPU, and nothing in this package uses it. Core catches only `ImportError` / `OSError` around that import, so a mismatched torchcodec's `RuntimeError` broke video input instead of falling back; without the package core reads with OpenCV. Removed in 6ed58ba7 together with the fork, the import test and the upper bound. The Windows DLL bootstrap stays for a torchcodec the user installs beside a matching torch, and the installation page describes that as optional, with the speed cost of the OpenCV path.

docs/get-started/installation.md
nghorbani: With torchcodec gone the reinstall line needs no third package. On `--index-url`: checked with `uv pip compile --python-platform aarch64-manylinux_2_28 --python-version 3.12` for `torch==2.11.0+cu128` and `torchvision==0.26.0+cu128`, once with `--index` (PyPI kept) and once with `--index-url` (PyPI replaced): identical resolutions, the cu128 index mirrors torch's whole dependency set (`cuda-toolkit`, `cuda-bindings`, the `nvidia-*` runtimes). The line stays as written.

tests/test_pyproject_torch_sources.py
nghorbani: The test left with the dependency (6ed58ba7). Its replacement, `test_torchcodec_is_not_a_dependency`, checks the dependencies, the extras and the lock; nothing in the suite imports torchcodec any more, so the pre-push gate no longer depends on FFmpeg shared libraries.

CHANGELOG.md
nghorbani: #106 is renumbered to 0.17.2 (title, description, changelog heading, builtin pin), so the two no longer collide; this one stays 0.17.1 and merges first, #106 rebases onto it. The date here is stamped in a last commit right before the squash, so it names the merge day.

nghorbani:
Thanks for the review. Folded in with 6ed58ba7 and 26c35a06: torchcodec removed instead of capped (your second question was the right one), the FFmpeg docs section rewritten (writer only; GPU decoding as an optional paragraph with the torch pairing), the `--index-url` point checked with a cross-platform `uv pip compile` (no change needed), the import test gone with the dependency, and #106 renumbered to 0.17.2. The Thor run stays deferred; the first command there is `uv run pytest tests/test_pyproject_torch_sources.py`. Re-requesting your review.

